PR #11, opened 2016-03-24
Isoform Override support

onursumer:
This PR adds support for isoform overrides:

- From the **/variant_annotation/isoform_override** service, isoform information for certain transcripts (from a specific override source) can be requested.
- **/variant_annotation/hgvs** now accepts an optional parameter **isoformOverrideSource**. When **isoformOverrideSource** is provided, this service takes isoform override information into account and enriches the returned JSON with the overridden canonical isoform information. Currently, enriched JSON is not saved back to the cache.

This service requires an additional line in application.properties, an example:
<code>vep.overrides_uris=uniprot=isoform_overrides_uniprot,mskcc=isoform_overrides_mskcc</code>

For an example isoform override file resource, see https://github.com/mskcc/vcf2maf/blob/master/data/isoform_overrides_uniprot

### DIFF
--- a/annotation/pom.xml
+++ b/annotation/pom.xml
@@ -58,6 +58,11 @@
       <version>2.3.1</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>com.univocity</groupId>
+      <artifactId>univocity-parsers</artifactId>
+      <version>2.0.0</version>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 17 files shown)


n1zea144:
Should we move the AnnotationEnricher and IsoformAnnotationEnricher into the service package?

onursumer:
We can also consider moving them into a new separate package. But for now, I think service package seems like a better place than domain package.

onursumer:
@n1zea144 @jjgao @zheins : do you think this PR is ready to merge?

zheins:
I've been using this branch in my development of the pipeline, seems to be working

jjgao:
I just merged it :)

onursumer:
Thanks jj!

Issue #18, opened 2016-11-01
Cancer hotspots


onursumer:
_From @jjgao on October 17, 2016 15:29_

- [ ] modeling cancer hotspots data in GN
- [ ] pull cancer htospots to annotate mutations

_Copied from original issue: cBioPortal/cbioportal#1814_

onursumer:
Basic algorithm:

- Allow users to Query Genome nexus by variant: /variant_annotation/cancer_hotspots/{variant}
- Get the `VariantAnnotation` for the provided variant
- Query www.cancerhotspots.org for each `TranscriptConsequence` in the `VariantAnnotation`
- If the `TranscriptConsequence` corresponds to a hotspot mutation, create a `Hotspot` instance and cache it.
- Return the list of `Hotspot`s as the result of the query.

onursumer:
Related PR: https://github.com/cBioPortal/cancerhotspots/pull/54

onursumer:
Updated the public cancer hotspots webservice. Now we are good to test this PR with the live cancer hotspots web service.

PR #23, opened 2016-11-14
Cancer Hotspots support

onursumer:
Added a new API (`/cancer_hotspots/{variants:.+}`) to query cancer hotspots web service.

This PR is intended to fix issue #18 

In order for this new API to work properly, `application.properties` should include a line like this:
`hotspots.url= http://cancerhotspots.org/api/hotspots/single/`

### DIFF
--- a/annotation/src/main/java/org/cbioportal/genome_nexus/annotation/util/Transformer.java
+++ b/annotation/src/main/java/org/cbioportal/genome_nexus/annotation/util/Transformer.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright (c) 2016 Memorial Sloan-Kettering Cancer Center.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY, WITHOUT EVEN THE IMPLIED WARRANTY OF MERCHANTABILITY OR FITNESS
+ * FOR A PARTICULAR PURPOSE. The software and documentation provided hereunder
+ * is on an "as is" basis, and Memorial Sloan-Kettering Cancer Center has no
+ * obligations to provide maintenance, support, updates, enhancements or
+ * modifications. In no event shall Memorial Sloan-Kettering Cancer Center be
+ * liable to any party for direct, indirect, special, incidental or
+ * consequential damages, including lost profits, arising out of the use of this
+ * software and its documentation, even if Memorial Sloan-Kettering Cancer
+ * Center has been advised of the possibility of such damage.
+ */
+
+/*
+ * This file is part of cBioPortal Genome Nexus.
+ *
+ * cBioPortal is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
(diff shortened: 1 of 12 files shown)


annotation/src/main/java/org/cbioportal/genome_nexus/annotation/util/Transformer.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * Maps the given raw JSON string onto the provided class instance.
+     *
+     * @param jsonString    raw JSON string
+     * @return a list of instances of the provided class
+     * @throws IOException
+     */
+    public static <T> List<T> mapJsonToInstance(String jsonString, Class<T> type) throws IOException
```
zheins: Spring has a ResponseEntity class that you can use to map rest api responses to java objects directly, you might want to look at it for handling responses from APIs. It might simplify some of this. I think there are other spots in the genome-nexus code that could use it as well.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -141,6 +148,56 @@ public AnnotationController(VariantAnnotationService variantAnnotationService,
        return getVariantAnnotation(variants, isoformOverrideSource);
     }
 
+    @ApiOperation(value = "getHotspotAnnotation",
```
zheins: Would it be a good idea to include hotspot annotation by default when getting variants? I'm thinking about how this is going to be used in the pipelines - we already make two calls to genome-nexus for the annotation and transcripts. Now we would have to make a third for the hotspot annotation. I'm sort of new to looking at this code, but I feel like combining the functionality of some of these endpoints might be a good idea.

annotation/src/main/resources/application.properties
```diff
@@ -1,3 +1,5 @@
 vep.url=http://grch37.rest.ensembl.org/vep/human/hgvs/VARIANT?content-type=application/json&xref_refseq=1&ccds=1&canonical=1&domains=1&hgvs=1&numbers=1&protein=1
+hotspots.url=http://cancerhotspots.org/api/hotspots/single/
+vep.overrides_uris=uniprot=isoform_overrides_uniprot.txt,mskcc=isoform_overrides_at_mskcc.txt
```
zheins: Kind of minor, but is it ok to use the properties field like this? If there are going to be multiple key-value pairs, maybe use ":" instead of "=", I feel like it looks cleaner that way.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/util/Transformer.java
onursumer: This is good to know. Thanks @zheins!

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
onursumer: @zheins: Good point. We may want to send a single query and get everything we need at once.

On the other hand, I think the ultimate goal is to avoid using genome-nexus in pipelines at all, but instead query data on the fly.

annotation/src/main/resources/application.properties
onursumer: @zheins sounds good to me. I shouldn't have included this property in this PR actually. It is part of the isoform support, not related to cancer hotspots. I will change it in the next isoform support PR.

zheins:
I also don't think application.properties should be checked in. Can you make an application.properties.EXAMPLE file, similar to the cbioportal repo? I lost my settings when I checked out this branch not realizing that file was part of the repo.

onursumer:
Agreed, it might be better to have an example file.

There was a `server.port=`field, which is used when building the jar version. We should add it to the example file, too.

annotation/src/main/resources/application.properties
onursumer: PR #27 now addresses this issue.

PR #27, opened 2016-11-17
Isoform Override API improvements

### LINKED ISSUE #21: return a list of available isoform sources in web api

onursumer:
- Added an API method to list all available isoform override data sources. (fixes #21)
- Updated isoform override uri resource delimiter. Now using semicolon instead of equals: `vep.overrides_uris=uniprot:isoform_overrides_uniprot.txt,mskcc:isoform_overrides_at_mskcc.txt`

### DIFF
--- a/annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
+++ b/annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
@@ -225,6 +225,32 @@ public class AnnotationController
         return isoformOverrideService.getIsoformOverrides(source);
     }
 
+    @ApiOperation(value = "getIsoformOverrideSources",
+        nickname = "getIsoformOverrideSources")
+    @ApiResponses(value = {
+        @ApiResponse(code = 200, message = "Success",
+            response = IsoformOverride.class,
+            responseContainer = "List"),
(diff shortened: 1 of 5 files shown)


annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -225,6 +225,32 @@ public AnnotationController(VariantAnnotationService variantAnnotationService,
         return isoformOverrideService.getIsoformOverrides(source);
     }
 
+    @ApiOperation(value = "getIsoformOverrideSources",
```
zheins: Minor, but maybe put a short sentence describing the endpoint like you do for the other ones.

onursumer: @zheins Good point. Actually we have a documentation problem here. Current descriptions are parameter level only, there is no method level description. Since this method doesn't have any parameters, we ended up with no description.

I think we should add method level description for all endpoints. It might be better to address this issue with another PR, since we will need to modify all endpoints. According to the [swagger documentation](https://github.com/swagger-api/swagger-core/wiki/Annotations#apioperation), we need to update the `value` field of each `ApiOperation` to add a short description.

Issue #31, opened 2016-11-21
Add description to API endpoint methods


onursumer:
Currently we only have parameter level description for the API methods in `AnnotationController`. Add a brief description for each public endpoint by editing the `value` field of the `ApiOperation` annotation.

Issue #28, opened 2016-11-18
Remove application.properties from the project


onursumer:
Replace `application.properties` with `application.properties.EXAMPLE`

PR #32, opened 2016-11-21
Documentation improvements

onursumer:
This PR addresses issues #28 and #31.


annotation/src/main/resources/application.properties.EXAMPLE
```diff
@@ -1,5 +1,6 @@
 vep.url=http://grch37.rest.ensembl.org/vep/human/hgvs/VARIANT?content-type=application/json&xref_refseq=1&ccds=1&canonical=1&domains=1&hgvs=1&numbers=1&protein=1
 hotspots.url=http://cancerhotspots.org/api/hotspots/single/
-vep.overrides_uris=uniprot=isoform_overrides_uniprot.txt,mskcc=isoform_overrides_at_mskcc.txt
+vep.overrides_uris=uniprot:isoform_overrides_uniprot.txt,mskcc:isoform_overrides_at_mskcc.txt
```
zheins: Why is this called "overrides_uris"?

annotation/src/main/resources/application.properties.EXAMPLE
```diff
@@ -1,5 +1,6 @@
...
 
-spring.data.mongodb.uri=
+spring.data.mongodb.uri=mongodb://127.0.0.1:27017/annotator
```
zheins: I think leaving this line blank is better, or replacing the address with `<HOST>`, `<PORT>`, and the db name to `<DB>`. Same thing for the `server.port` value.

zheins:
Also, I think squashing the commits together into one is a good idea.

annotation/src/main/resources/application.properties.EXAMPLE
onursumer: I was thinking the possibility of getting the content from an external URL such as https://raw.githubusercontent.com/mskcc/vcf2maf/master/data/isoform_overrides_at_mskcc. But maybe `override_sources` would be a better naming choice.

annotation/src/main/resources/application.properties.EXAMPLE
onursumer: Since this is an example file, I think it is okay to have a working example value as long as it is not confidential. We can add a comment on top of each property to make things clearer.

@zheins What do you think?

onursumer:
@zheins I am planning to use the "squash and merge" feature once the PR is ready to merge.

annotation/src/main/resources/application.properties.EXAMPLE
```diff
@@ -1,6 +1,16 @@
...
 vep.url=http://grch37.rest.ensembl.org/vep/human/hgvs/VARIANT?content-type=application/json&xref_refseq=1&ccds=1&canonical=1&domains=1&hgvs=1&numbers=1&protein=1
+
+# Cancer Hotspots web API URL
 hotspots.url=http://cancerhotspots.org/api/hotspots/single/
+
+# Comma separated list of isoform_overrides_resource:overrides_filename pairs.
 vep.overrides_uris=uniprot:isoform_overrides_uniprot.txt,mskcc:isoform_overrides_at_mskcc.txt
 
+# MongoDB URI in the form of mongodb://<USERNAME>:<PASSWORD>@<HOST>:<PORT>/<DB>
 spring.data.mongodb.uri=mongodb://127.0.0.1:27017/annotator
```
zheins: Should there be a real value in this field? or just use what you have in the above comment here? Same comment for port field.

annotation/src/main/resources/application.properties.EXAMPLE
zheins: Yeah I agree, go ahead and change the name, I think it makes more sense.

annotation/src/main/resources/application.properties.EXAMPLE
zheins: Sounds good!

Issue #29, opened 2016-11-18
Add support for combining the functionality of some endpoints


onursumer:
Currently we need to query the API several times for different kind of information. Adding support to combine information from multiple data sources into one API would be nice.

For example: By a single query, get the VEP annotation for a specific isofom override together with the cancer hotspot information.

onursumer:
Completed for the current endpoints. We may need to revisit this issue once we add more endpoints. Closing for now.

PR #35, opened 2016-11-22
Optional cancer hotspots information for VEP annotation

onursumer:
- added an enricher to add hotspot info into transcripts
- added an API parameter to optionally enable hotspot enrichment

example query: http://localhost/variant_annotation/hgvs/7:g.140453136A>T?isoformOverrideSource=uniprot&cancerHotspots=summary

Intended for the issue #29

### DIFF
--- a/annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
+++ b/annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
@@ -34,11 +34,11 @@ package org.cbioportal.genome_nexus.annotation.web;
 
 import io.swagger.annotations.*;
 import org.cbioportal.genome_nexus.annotation.domain.*;
+import org.cbioportal.genome_nexus.annotation.service.internal.HotspotAnnotationEnricher;
 import org.cbioportal.genome_nexus.annotation.service.internal.IsoformAnnotationEnricher;
 import org.cbioportal.genome_nexus.annotation.service.*;
 
 import org.cbioportal.genome_nexus.annotation.service.internal.VEPEnrichmentService;
-import org.cbioportal.genome_nexus.annotation.util.Numerical;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.client.HttpClientErrorException;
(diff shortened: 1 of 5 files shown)


annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -314,29 +335,14 @@ private IsoformOverride getIsoformOverride(String source, String transcriptId)
...
+        List<Hotspot> hotspots = hotspotService.getHotspots(transcript);
 
-                hotspots.add(hotspot);
-
-                // do not cache anything for now
-                //hotspotRepository.save(hotspots);
-            }
-        }
+        // do not cache anything for now
+        //hotspotRepository.save(hotspots);
```
zheins: Whats the reason for this?

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -114,6 +120,15 @@ public AnnotationController(VariantAnnotationService variantAnnotationService,
             postEnrichmentService.registerEnricher(isoformOverrideSource, enricher);
         }
 
+        // ignore any other value than "full" or "summary"
+        if (cancerHotspots != null &&
+            (cancerHotspots.equalsIgnoreCase("full") || cancerHotspots.equalsIgnoreCase("summary")))
```
zheins: A more general comment about the code, I notice a lot of hardcoded strings. Maybe some of these values could be final strings in some class or an enum? Probably shouldn't be addressed in this PR

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
onursumer: Initially we were planning to query the cancerhotspots API for each transcript separately, and then cache the result in the genome-nexus database, but later we decided to retrieve all hotspot mutations at once and cache them in memory. So, currently the hotspot repository is not used.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
onursumer: Agreed. It is generally a good idea to have final strings or enums instead of hardcoded strings.

PR #40, opened 2017-06-14
Support few more properties in TranscriptConsequence

zhx828:
These properties are not part of VEP format, but are useful information
for other purposes. They will only be initialized when user select
DETAILED projection


zhx828:
@onursumer this pull request needs to be tested with MongoDB.

onursumer:
We may want to revisit the `DETAILED` projection idea after merging #172, probably in a new PR though.

inodb:
gonna close because of inactivity

PR #41, opened 2017-06-14
Mutation assessor [WIP]

athum:
Add Mutation Assessor endpoint [Work in Progress]


annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/MutationAssessorService.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.io.IOException;
+import java.util.List;
+
+public class MutationAssessorService
+{
+
+    private static String mutationAssessorURL;
+
+    public void setMutationAssessorURL(String mutationAssessorURL)
```
onursumer: We should get this value from `application.proerties`. Easiest way to do it is to add an annotation for the setter, something like this: `@Value("${mutationAssessor.url}")`

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/MutationAssessorService.java
```diff
@@ -0,0 +1,79 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.web.client.RestTemplate;
+import com.mongodb.DBObject;
+import com.mongodb.util.JSON;
+import org.cbioportal.genome_nexus.annotation.domain.MutationAssessor;
+
+import java.io.IOException;
+import java.util.List;
+
+public class MutationAssessorService
```
onursumer: We need to add `@Service` annotation for the class

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -52,27 +53,30 @@
...
-@RequestMapping(value = "/")
+@RequestMapping(value= "/")
 public class AnnotationController
 {
     private final VariantAnnotationService variantAnnotationService;
     private final VariantAnnotationRepository variantAnnotationRepository;
     private final IsoformOverrideService isoformOverrideService;
     private final HotspotService hotspotService;
     private final HotspotRepository hotspotRepository;
+    private final MutationAssessorService mutationService;
```
onursumer: Let's name this variable `mutationAssessorService` to avoid any confusion.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -218,6 +222,64 @@ public AnnotationController(VariantAnnotationService variantAnnotationService,
...
     }
 
+    /*
+        *
+        *
+        *
+        *
+        *
+        *
+        * */
```
onursumer: I don't think we need empty comment blocks :)

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/MutationAssessorTest.java
```diff
@@ -0,0 +1,35 @@
+package org.cbioportal.genome_nexus.annotation.service.internal;
+
+import org.cbioportal.genome_nexus.annotation.domain.MutationAssessor;
+import org.cbioportal.genome_nexus.annotation.service.internal.MutationAssessorService;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+public class MutationAssessorTest {
```
onursumer: We need to remove this class...

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -218,6 +222,64 @@ public AnnotationController(VariantAnnotationService variantAnnotationService,
...
+    @ApiResponses(value = {
+        @ApiResponse(code = 200, message = "Success",
+            response = MutationAssessor.class,
+            responseContainer = "List"),
+        @ApiResponse(code = 400, message = "Bad Request")
+    })
+    @RequestMapping(value = "/mutation_assessor/{variants:.+}",
+        method = RequestMethod.GET,
+        produces = "application/json")
+    public List<MutationAssessor> getMutationAnnotation(
```
onursumer: Let's rename this one to `getMutationAssessorAnnotation`

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/domain/MutationAssessor.java
```diff
@@ -0,0 +1,70 @@
...
+ @JsonInclude(JsonInclude.Include.NON_NULL)
+    public class MutationAssessor
+    {
+
+        // user-input variant
+        private String variant;
+
+        private String hugoSymbol;
+        private String functionalImpact;
+        private String fScore;
```
onursumer: I think this one should be a number, not a string. Also `functionalImpactScore` might be a better variable name than `fScore`.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/domain/MutationAssessor.java
```diff
@@ -0,0 +1,70 @@
...
+            return this.fScore;
+        }
+
+        @JsonProperty("F_score")
+        public void setfScore(String fScore)
+        {
+            this.fScore = fScore;
+        }
+
+        public String getOuput()
```
onursumer: I guess this is for debugging purposes

annotation/src/test/java/org/cbioportal/genmone_nexus/annotation/util/MutationAssessorTest.java
```diff
@@ -0,0 +1,98 @@
...
+{
+    /* for debugging
+    private static Logger log = Logger.getLogger(String.valueOf(MutationAssessorTest.class));
+    */
+
+    @Test
+    public void testVariants() throws IOException {
+
+        MutationAssessorService service = new MutationAssessorService();
+        service.setMutationAssessorURL("http://mutationassessor.org/r3/?cm=var&var=");
```
onursumer: I am not sure if it is a good idea to use an actual/live web API for testing purposes. It is not a reliable way of testing. The common practice is to create a mock data/api. Unfortunately there is no example test in genome nexus code base. Maybe we can find some examples in cBioPortal API classes.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -337,14 +385,16 @@ private IsoformOverride getIsoformOverride(String source, String transcriptId)
 
     private List<Hotspot> getHotspotAnnotation(TranscriptConsequence transcript)
     {
-        //String transcriptId = transcript.getTranscriptId();
-        //Hotspot hotspot = hotspotRepository.findOne(transcriptId);
 
+        // String transcriptId = transcript.getTranscriptId();
+        // Hotspot hotspot = hotspotRepository.findOne(transcriptId);
+
+        // hotspotService.setHotspotsURL("http://cancerhotspots.org/api/hotspots/single/");
```
onursumer: Ideally there should not be any code change in this PR related to hotspot annotation. Is it possible to revert `getHotspotAnnotation` content to its original version for now?

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -218,6 +225,47 @@ public AnnotationController(VariantAnnotationService variantAnnotationService,
...
+        nickname = "postMutationAssessorAnnotation")
+    @RequestMapping(value = "/mutation_assessor",
+        method = RequestMethod.POST,
+        produces = "application/json")
+    public List<MutationAssessor> postMutationAssessorAnnotation(
+        @RequestParam
+        @ApiParam(value="Comma separated list of variants. For example 7:g.140453136A>T,12:g.25398285C>A",
+            required = true,
+            allowMultiple = true)
+        List<String> variants)
```
onursumer: A general concern: What we have been doing so far with the `list parameters` is not optimal. We should try to avoid comma separated query parameters, and replace all existing ones with body parameters. We don't need to worry about it for now though, it is already like this everywhere :)

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/MutationAssessorService.java
```diff
@@ -0,0 +1,72 @@
...
+
+        // todo: check that variant is in the right format
+        if (variants != null &&
+            variants.length() > 0)
+        {
+            uri += variants + "&frm=json";
+        }
+
+        RestTemplate restTemplate = new RestTemplate();
+        return restTemplate.getForObject(uri, String.class);
```
onursumer: Does Mutation Assessor web service support POST instead of GET? For a long list of variants this query may fail. In case the variant list is too long we need to use either a POST request or multiple GET requests.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/MutationAssessorService.java
```diff
@@ -0,0 +1,72 @@
...
+
+        // todo: handle this
+        return null;
+    }
+
+
+    // todo: standardize input formatting
+    private static String toMutationString(String inputString)
+    {
+        String temp;
```
onursumer: I think this method deserves some comment/explanation.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
inodb: @onursumer Do you know of an example of this elsewhere? Maybe in `cbioportal` repo? And what exactly is suboptimal about using `list parameters`?

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
athum: Nothing was actually modified, I just uncommented and re-commented some code when I was testing it out

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
onursumer: Using a list parameter is okay. But getting the list as a comma separated `query` or `path` parameter is suboptimal. We need to change all of these to `body` parameters at some point. A very long `path` or `query` parameter may potentially break the query. AFAIK, in cBioPortal API, list parameters are defined as `body` parameters.

annotation/src/test/java/org/cbioportal/genmone_nexus/annotation/util/MutationAssessorTest.java
inodb: @onursumer Yeah I told @athum to not worry about that for now. I can help address this issue in a separate PR.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/MutationAssessorService.java
athum: There's no POST method specified in the API, so for now I've just been iterating through the variant list and making separate GET requests.

PR #46, opened 2017-06-28
Fixed logging to prevent bloating server logs

angelicaochoa:
Fixed logging to prevent bloating logs when `DataIntegrityViolationException`'s are thrown

Added try/catch when saving a variant annotation json to mongodb. This is to prevent bloating the logs by large stack traces when a `DataIntegrityViolationException` is thrown - now only the localized error message is logged.

### DIFF
--- a/annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
+++ b/annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
@@ -42,10 +42,11 @@ import org.cbioportal.genome_nexus.annotation.service.internal.VEPEnrichmentServ
 import org.springframework.web.bind.annotation.*;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.client.HttpClientErrorException;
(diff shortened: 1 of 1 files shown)


annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -370,7 +373,16 @@ private VariantAnnotation getVariantAnnotation(String variant)
                 variantAnnotation = variantAnnotationRepository.mapAnnotationJson(variant, annotationJSON);
 
                 // save everything to the cache as a properly parsed JSON
-                variantAnnotationRepository.saveAnnotationJson(variant, annotationJSON);
+                
+                try {
+                    variantAnnotationRepository.saveAnnotationJson(variant, annotationJSON);
+                }
+                catch (DataIntegrityViolationException e) {
```
onursumer: @angelicaochoa I would add this `catch` block for the outer `try` to avoid nested try statements... Unless we really want to catch this exception only for `saveAnnotationJson` method, and we don't want to catch it for the rest of the block.

PR #47, opened 2017-07-12
Fixed VariantAnnotation object returned when  DataIntegrityViolationException thrown

angelicaochoa:
If a `DataIntegrityViolationException` is thrown then the `VariantAnnotation` object returned will be malformed because of the additional caught `HttpClientErrorException`, which will overwrite the annotationJSON used to construct the `VariantAnnotation` object. This makes Genome Nexus think that it failed to annotate a variant even when the hgvs variant used is valid and returns a valid response from the VEP server.

If the hgvs variant is valid and returns a valid response from VEP server then we still want to parse the annotationJSON into a`VariantAnnotation` object.

Instead of overwriting the annotationJSON with the `HttpClientErrorException` message, the `VariantAnnotation` object will now be returned early to prevent the overwrite from happening.

### DIFF
--- a/annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
+++ b/annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
@@ -382,6 +382,7 @@ public class AnnotationController
                     // this is thrown when the annotationJSON can't be stored by mongo
                     // due to the variant annotation key being too large to index
                     LOG.info(e.getLocalizedMessage());
+                    return variantAnnotation;
                 }
             }
             catch (HttpClientErrorException e) {


annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -382,6 +382,7 @@ private VariantAnnotation getVariantAnnotation(String variant)
                     // this is thrown when the annotationJSON can't be stored by mongo
                     // due to the variant annotation key being too large to index
                     LOG.info(e.getLocalizedMessage());
+                    return variantAnnotation;
```
sheridancbio: I don't see how this would change behavior. If the flow of control reached this point and there was no return statement here, it seems that we have consumed the DataIntegrityViolationException and would next exit the  outer try block (line 362 - 387) without further exception, so we would skip the two catch blocks below and reach the (identical) return on line 400. Maybe somehow the change to variantAnnotation within the try blocks (line 373) is being undone because of unwinding the inner try block during the DataIntegrityViolation Exception?

zheins: I agree with Rob on this. I have an instance of genome nexus running master here: http://dashi-dev.cbio.mskcc.org:5556. I tried the links you provided in slack and it returns the correct annotation.

zheins: To me, this looks like maybe there is a logging configuration issue when the app runs on tomcat. Can you explore this? My instance on dashi-dev is running as a jar on the command line. Maybe it does not log correctly when running on a tomcat.

zheins:
Issue resolved by redeploying master to genomenexus.org. I think this PR can be closed

angelicaochoa:
Closing this PR - the issue was that http://genomenexus.org/ had not been updated with latest changes to master which addressed this issue

PR #50, opened 2017-08-09
Expanded mutation assessor response model

athum:
Expanded mutation assessor response model


sheridancbio:
I think this is really great to see. Hopefully we can soon strip the mutationassessor fields from the mutation_event table.

One comment though: when it comes time to merge the mutation-assessor branch into master I hope that we first squash all of these commits into a single commit. The fact that some of these commits are merge commits may make that a pain when the time comes.

PR #56, opened 2017-08-28
Update swagger version

inodb:
For generation of typescript client in cbioportal/cbioportal-frontend it is
essential to have a newer version of Swagger, otherwise all objects returned by
the API have `any` type.


inodb:
@sheridancbio updated to 2.7.0

PR #62, opened 2017-09-21
New Endpoint for PFAM domains

### LINKED ISSUE #60: Gene to PFAM domain mapping
Use Ensembl biomart to get Gene -> Transcript -> Translation -> PFAM domain mappings: http://www.ensembl.org/biomart/martview/

We need to generate a TSV file from this website and the use it to build our DB. See the screenshot for the required attributes when constructing the TSV file.

biomart_tsv_export

onursumer:
Fixes #60

Added a new endpoint for PFAM domains. The `PfamDomain` model currently reflects the biomart file, but we should probably extend it with additional domain information.

### DIFF
--- a/annotation/src/main/resources/application.properties.EXAMPLE
+++ b/annotation/src/main/resources/application.properties.EXAMPLE
@@ -15,6 +15,9 @@ mutationAssessor.url=http://mutationassessor.org/r3/?cm=var&var=VARIANT&frm=json
 # Comma separated list of isoform_overrides_resource:overrides_filename pairs.
 vep.isoform.overrides=uniprot:isoform_overrides_uniprot.txt,mskcc:isoform_overrides_at_mskcc.txt
 
+# Ensemble biomart PFAM filename
+ensembl.biomart.pfam=ensembl_biomart_pfam.txt
+
 # MongoDB URI in the form of mongodb://<USERNAME>:<PASSWORD>@<HOST>:<PORT>/<DB>
 spring.data.mongodb.uri=mongodb://127.0.0.1:27017/annotator
 
--- a/README.md
+++ b/README.md
@@ -25,3 +25,17 @@ for interpreting variants and patients.
 mvn clean install
 java -jar annotation/target/annotation-*.war
 ```
+
+## Data Download
+
+### Ensemble Biomart
+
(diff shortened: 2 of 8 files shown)


annotation/src/main/resources/application.properties.EXAMPLE
```diff
@@ -15,6 +15,9 @@ mutationAssessor.url=http://mutationassessor.org/r3/?cm=var&var=VARIANT&frm=json
 # Comma separated list of isoform_overrides_resource:overrides_filename pairs.
 vep.isoform.overrides=uniprot:isoform_overrides_uniprot.txt,mskcc:isoform_overrides_at_mskcc.txt
 
+# Ensemble biomart PFAM filename
+ensemble.biomart.pfam=ensembl_biomart_pfam.txt
```
inodb: @onursumer should add this fail to the PR

onursumer: @inodb We don't have `isoform_overrides_uniprot.txt`, `isoform_overrides_at_mskcc.txt` in the repository either. But I guess it makes sense to maintain the biomart file in this repository.

inodb: @onursumer good point. I think we should actually store those and have some script to update/download them. Check if they're in sync. I'll create an issue

inodb:
@onursumer could we change the property to default to the included biomart file path `ensembl.biomart.pfam`. Right now heroku fails because it can't find the biomart file. I think we should set up in such a way, such that it builds without doing any extra steps.

```
2017-09-26T09:14:31.995875+00:00 app[web.1]: java.lang.IllegalArgumentException: Could not resolve placeholder 'ensembl.biomart.pfam' in string value "${ensembl.biomart.pfam}"
2017-09-26T09:14:31.995876+00:00 app[web.1]:    at org.springframework.util.PropertyPlaceholderHelper.parseStringValue(PropertyPlaceholderHelper.java:174) ~[spring-core-4.2.4.RELEASE.jar!/:4.2.4.RELEASE]
```

onursumer:
@inodb There was a typo in `application.properties.EXAMPLE`, now the heroku instance should work fine.

PR #65, opened 2017-09-27
Added name and description info for PFAM domains

### LINKED ISSUE #61: Get PFAM domain name, description, and color
Find a service to get the domain name, description and color for a given PFAM domain id (such as PF00146, PF06444, etc).

PFAM database might be useful for this: http://pfam.xfam.org/help#tabview=tab12

onursumer:
- Fixed #61
- Also updated PFAM endpoint URLs (see https://stackoverflow.com/q/20381976 for details)
- We decided to not include color mappings in genome-nexus, we will dynamically assign colors on the front end.

### DIFF
--- a/annotation/src/main/resources/application.properties.EXAMPLE
+++ b/annotation/src/main/resources/application.properties.EXAMPLE
@@ -18,6 +18,9 @@ vep.isoform.overrides=uniprot:isoform_overrides_uniprot.txt,mskcc:isoform_overri
 # Ensemble biomart PFAM filename
 ensembl.biomart.pfam=ensembl_biomart_pfam.txt
 
+# PFAM_A filename
+pfam.pfamA=pfamA.txt
+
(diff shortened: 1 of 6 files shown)


annotation/src/main/resources/application.properties.EXAMPLE
```diff
@@ -18,6 +18,9 @@ vep.isoform.overrides=uniprot:isoform_overrides_uniprot.txt,mskcc:isoform_overri
 # Ensemble biomart PFAM filename
 ensembl.biomart.pfam=ensembl_biomart_pfam.txt
 
+# PFAM_A filename
+pfam.pfamA=pfamA.txt
```
inodb: @onursumer maybe include this file in the PR?

onursumer: @inodb the full file is about 22MB, but actually we parse only 3 columns for now. Maybe we should only include a truncated version of the file?

onursumer: decided to include those 3 columns only. If we need more columns from the full file later, we can update our file.

PR #68, opened 2017-10-05
Improvements for AnnotationController, and for the software architecture

onursumer:
# Changes

- Moved xref endpoint to a separate controller
- Removed all endpoints except `annotation` from the swagger documentation
- Some of the removed endpoints are still available. Keeping those hidden endpoints for now to avoid breaking our internal pipelines.
- Added more controllers for internal API endpoints, and added an internal API group
- Minor refactoring of the business logic: Moved certain methods from controller and persistence layers into the service layer, and simplified related interfaces.
- Introduced mixing to cleanup models and group model annotations into separate layers: For example annotations for API layer is handled by mixins under web package, and for service layer by mixins under service package.

# Still Need To Improve

- Annotations for persistence layer (MongoDB and file parsing annotations) still remain within the actual model classes. Ideally we should also be able to handle that configuration outside the model classes.
- Some endpoints currently return empty result with status code 200 in case no record is found. Ideally we should return `404 Not Found` for those queries.
- More tests, especially for the business logic.


inodb:
@onursumer what do you think about using groups instead, similar to how cBioPortal API works. (We could make a similar `group=internal`): http://www.cbioportal.org/api/swagger-ui.html. Didn't think of this before, perhaps better if we want to continue to use the docs ourselves to explore the currently existing APIs.

onursumer:
@inodb sounds good to me. I can update the PR, so that we can add a controller for the internal API

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -155,41 +168,42 @@ public AnnotationController(VariantAnnotationService variantAnnotationService,
...
-            response = Hotspot.class,
-            responseContainer = "List"),
-        @ApiResponse(code = 400, message = "Bad Request")
-    })
-    @RequestMapping(value = "/cancer_hotspots/{variants:.+}",
-        method = RequestMethod.GET,
-        produces = "application/json")
+//    @ApiOperation(value = "Retrieves hotspot annotation for the provided list of variants",
+//        nickname = "getHotspotAnnotation")
+//    @RequestMapping(value = "/cancer_hotspots/{variants:.+}",
```
sheridancbio: If there we are dropping the /cancer_hotspots endpoint, shouldn't we move the method "getHotspotAnnotation()" out of the controller and into a service layer? Also, if the function is not mapped to an endpoint with RequestMapping, I wonder what the other java annotations (\@PathVariable \@ApiParam) do? Maybe there is still a default endpoint mapping for this which is not requested?

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -229,17 +243,11 @@ public AnnotationController(VariantAnnotationService variantAnnotationService,
...
-    @ApiResponses(value = {
-        @ApiResponse(code = 200, message = "Success",
-            response = MutationAssessor.class,
-            responseContainer = "List"),
-        @ApiResponse(code = 400, message = "Bad Request")
-    })
-    @RequestMapping(value = "/mutation_assessor/{variants:.+}",
-        method = RequestMethod.GET,
-        produces = "application/json")
+//    @ApiOperation(value = "Retrieves mutation assessor information for the provided list of variants",
```
sheridancbio: same comment as previous

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -275,17 +283,12 @@ public AnnotationController(VariantAnnotationService variantAnnotationService,
...
-            responseContainer = "List"),
-        @ApiResponse(code = 400, message = "Bad Request")
-    })
-    @RequestMapping(value = "/mutation_assessor",
-        method = RequestMethod.POST,
-        produces = "application/json")
+//    @ApiOperation(value = "Retrieves mutation assessor information for the provided list of variants",
+//        hidden = true,
+//        nickname = "postMutationAssessorAnnotation")
+//    @RequestMapping(value = "/mutation_assessor",
```
sheridancbio: same comment as previous

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -398,16 +389,6 @@ public AnnotationController(VariantAnnotationService variantAnnotationService,
         return isoformOverrideService.getOverrideSources();
     }
 
-    @ApiOperation(value = "Gets a list of available isoform override data sources",
-        nickname = "postIsoformOverrideSources")
-    @RequestMapping(value = "/isoform_override/sources",
```
sheridancbio: same comment as previous (maybe this should become a service if it is not an endpoint)

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
```diff
@@ -111,35 +103,56 @@ public AnnotationController(VariantAnnotationService variantAnnotationService,
...
+        @ApiParam(value="Isoform override source. For example uniprot",
+            required = false)
+            String isoformOverrideSource,
+        @RequestParam(required = false)
+        @ApiParam(value="Comma separated list of fields to include (case-sensitive!). " +
+            "For example: hotspots,mutation_assessor", required = false, defaultValue = "hotspots,mutation_assessor")
+            List<String> fields)
+    {
+		List<VariantAnnotation> variantAnnotations = new ArrayList<>();
+        EnrichmentService postEnrichmentService = this.initPostEnrichmentService(isoformOverrideSource, fields);
```
sheridancbio: I don't fully understand this pattern - if *fields* comes in with every request and is important for every enrichAnnotation() call, then why not pass fields in as an argument to the service instead of instantiating a new instance of the service with the fields passed in as a constructor argument? This seems like a waste of heap to me; but I'm sure there is  reason (maybe some inheritance / polymorphism issue). We could still hold the service instances in a cache, so we don't need to keep instantiating them perhaps.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
onursumer: Actually, I am working on a separate internal `CancerHotspotsController` for this endpoint.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
onursumer: A separate internal `MutationAssessorController` is on the way.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
onursumer: A separate internal `IsoformOverrideController` is on the way.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/AnnotationController.java
onursumer: Oh, this one is a bit tricky. Actually, how we enrich the core VEP annotaiton depends on the `fields`. With this approach we re-instantiate the service and register enrichers for each request. It is like re-configuring the enrichment service for each query separately. But, I think we can optimize this way we configure the enrichment per query. It may not be necessary to re-instantiate the service each time.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/VEPVariantAnnotationService.java
```diff
@@ -32,40 +32,118 @@
...
-    public String getRawAnnotation(String variant)
+    /**
+     * Maps the given raw annotation JSON string onto a VariantAnnotation instance.
+     *
+     * @param variant           variant key
+     * @param annotationJSON    raw annotation JSON string
+     * @return a VariantAnnotation instance
+     * @throws IOException
+     */
+    private VariantAnnotation mapAnnotationJson(String variant, String annotationJSON) throws IOException
```
onursumer: Moved here from the persistence layer (`VariantAnnotationRepositoryImpl`)

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/VEPVariantAnnotationService.java
```diff
@@ -32,40 +32,118 @@
...
 
     public VariantAnnotation getAnnotation(String variant)
     {
-        String vepJSON = this.getRawAnnotation(variant);
-        VariantAnnotation variantAnnotation = new VariantAnnotation(variant, vepJSON);
-        enrichmentService.enrichAnnotation(variantAnnotation);
+        VariantAnnotation variantAnnotation = variantAnnotationRepository.findOne(variant);
+        String annotationJSON = null;
+
+        if (variantAnnotation == null) {
```
onursumer: Moved from the controller layer (`AnnotationController`)

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/VEPVariantAnnotationService.java
```diff
@@ -32,40 +32,118 @@
...
+     * @param variant           variant key
+     * @param annotationJSON    raw annotation JSON string
+     * @return a VariantAnnotation instance
+     * @throws IOException
+     */
+    private VariantAnnotation mapAnnotationJson(String variant, String annotationJSON) throws IOException
+    {
+        // map annotation string onto VariantAnnotation instance
+        List<VariantAnnotation> list = Transformer.mapJsonToInstance(
+            annotationJSON, VariantAnnotation.class, new ExternalSourceObjectMapper());
```
ersinciftci: Would be better if those Object Mappers were singleton beans, instead of creating a new one every time with `new`.

sheridancbio:
the (now hidden) hgvs endpoint still provides backwards compatibility, returning a list of objects as before. Merging this PR now, as it is been put into production.

Additional improvements can follow in future PRs

PR #70, opened 2017-10-17
Add Ensembl endpoints

inodb:
Add Python scripts to download tables from Ensembl. Perform data
transformations and load data in Genome Nexus to serve over ensembl endpoints.
Importing data first in mongodb on init.

### cBioPortal Usages:
- We want to replace the current pfam service to use genome nexus instead.
- Display appropriate length of protein in mutation diagram based on chosen isoform override (this will be a new property in cbioportal/cbioportal)


annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/EnsemblController.java
```diff
@@ -0,0 +1,187 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@RestController // shorthand for @Controller, @ResponseBody
+@CrossOrigin(origins="*") // allow all cross-domain requests
+@RequestMapping(value= "/")
+@Api(tags = "ensembl-controller", description = "Ensembl Controller")
```
onursumer: Also add `@PublicApi` annotation?

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/domain/EnsemblTranscript.java
```diff
@@ -0,0 +1,66 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.univocity.parsers.annotations.Parsed;
+import com.univocity.parsers.annotations.Trim;
+import io.swagger.annotations.ApiModelProperty;
+
+/**
+ * Ensembl transcript
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
onursumer: Ideally we should add API related annotations like `@JsonInclude` and `@ApiModelProperty` in a mixin under web package, but we can also do that in another PR.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/GenomeNexusAnnotation.java
```diff
@@ -55,8 +61,14 @@
  */
 @SpringBootApplication // shorthand for @Configuration, @EnableAutoConfiguration, @ComponentScan
 @EnableSwagger2 // enable swagger2 documentation
-public class GenomeNexusAnnotation extends SpringBootServletInitializer
+public class GenomeNexusAnnotation extends SpringBootServletInitializer implements CommandLineRunner
```
onursumer: Would it be too difficult to implement the `CommandLineRunner` in a separate class? Something like `class GenomeNexusImporter implements CommandLineRunner`.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/GenomeNexusAnnotation.java
```diff
@@ -102,4 +114,85 @@ private ApiInfo apiInfo() {
...
     }
+
+    /**
+     * Export a resource embedded into a Jar file to the local file path.
+     *
+     * @param resourceName ie.: "/SmartLibrary.dll"
+     * @return The path to the exported resource
+     * @throws Exception
+     */
+    static public String ExportResource(String resourceName) throws Exception {
```
onursumer: It is against Java coding conventions to have a method name starting with an uppercase letter :) Can we rename it to `exportResource`?

inodb: @onursumer ah yes, this function was a copy+paste from SO. Guess this was some C++/C# dev: https://www.reddit.com/r/java/comments/3cu10n/do_any_of_you_start_static_methods_with_a_capital/

inodb:
@onursumer ok should be good now

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/EnsemblController.java
inodb: We decided to go for internal instead, because we're not 100% happy about the endpoints yet (specifically isoform override)

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/domain/EnsemblTranscript.java
```diff
@@ -0,0 +1,59 @@
...
+    private String transcriptId;
+
+    @Field(value="gene_stable_id")
+    private String geneId;
+
+    @Field(value="protein_stable_id")
+    private String proteinId;
+
+    @Field(value="protein_length")
+    private String proteinLength;
```
onursumer: Better to make this an `Integer`

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/web/mixin/EnsemblTranscriptMixin.java
```diff
@@ -0,0 +1,19 @@
...
+    private String transcriptId;
+
+    @ApiModelProperty(value = "Ensembl gene id", required = true)
+    private String geneId;
+
+    @ApiModelProperty(value = "Ensembl protein id", required = true)
+    private String proteinId;
+
+    @ApiModelProperty(value = "Length of protein")
+    private String proteinLength;
```
onursumer: This one needs to change to `Integer` too

onursumer:
@inodb I think this is good to merge, except the type of `proteinLength`

PR #74, opened 2017-10-24
Divided annotation module into multiple modules

onursumer:
There are still some dependency issues like:
- Multiple modules still depend on the module `spring-boot-starter-web`. Ideally only `web` module should depend on it.
- `Model` module depends on `spring-boot-starter-data-mongodb`. Ideally only persistence module should depend on it.


onursumer:
@inodb Seems like `src/main/java/org/cbioportal/genome_nexus/annotation/util/parse/HugoSymbolRecord.java` is now obsolete. It should be okay to remove it, right?

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/AnnotationEnricher.java
```diff
@@ -1,11 +0,0 @@
-package org.cbioportal.genome_nexus.annotation.service;
-
-import org.cbioportal.genome_nexus.annotation.domain.VariantAnnotation;
-
-/**
- * @author Selcuk Onur Sumer
- */
-public interface AnnotationEnricher
```
onursumer: This file is not removed, just renamed, for some reason github shows it removed...

model/src/main/java/org/cbioportal/genome_nexus/model/EnsemblTranscript.java
```diff
@@ -1,13 +1,13 @@
...
 
-import org.cbioportal.genome_nexus.annotation.domain.internal.EnsemblRepositoryImpl;
 import org.springframework.data.mongodb.core.mapping.Document;
 import org.springframework.data.mongodb.core.mapping.Field;
 
 /**
  * Ensembl transcript
  */
-@Document(collection=EnsemblRepositoryImpl.TRANSCRIPTS_COLLECTION)
+//@Document(collection=EnsemblRepositoryImpl.TRANSCRIPTS_COLLECTION)
```
onursumer: Commented this out to prevent `model` module to depend on `persistence` module. We may want to keep these kind of shared constants in a different module rather than inside a specific `Impl` class.

util/src/test/java/util/NumericalTest.java
```diff
@@ -30,7 +30,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-package org.cbioportal.genmone_nexus.annotation.util;
+package util;
```
onursumer: Seems like a minor refactoring issue, might be better to keep this class in the `org.cbioportal.genmone_nexus.util` package instead of just `util`

web/src/main/java/org/cbioportal/genome_nexus/GenomeNexusAnnotation.java
```diff
@@ -47,18 +47,16 @@
...
 
-import java.io.*;
 import java.util.Arrays;
 import java.util.HashSet;
 
 /**
  * @author Benjamin Gross
  */
-@SpringBootApplication // shorthand for @Configuration, @EnableAutoConfiguration, @ComponentScan
+@SpringBootApplication(scanBasePackages = {"org.cbioportal.genome_nexus"}) // shorthand for @Configuration, @EnableAutoConfiguration, @ComponentScan
```
onursumer: `scanBasePackages = {"org.cbioportal.genome_nexus"}` required due to multiple modules

model/pom.xml
```diff
@@ -0,0 +1,34 @@
...
+    <parent>
+        <groupId>org.cbioportal.genome_nexus</groupId>
+        <artifactId>genome-nexus</artifactId>
+        <version>0.1.0</version>
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter</artifactId>
```
ersinciftci: `spring-boot-starter-data-mongodb` should already have this dependency, you can probably remove it.

persistence/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.cbioportal.genome_nexus</groupId>
+            <artifactId>model</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter</artifactId>
```
ersinciftci: spring-boot-starter-data-mongodb should already have this dependency, you can probably remove it.

persistence/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-data-mongodb</artifactId>
```
ersinciftci: Model already depends on this, and this module depends on Model, so you can remove this dependency.

persistence/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.univocity</groupId>
+            <artifactId>univocity-parsers</artifactId>
```
ersinciftci: Model already depends on this, and this module depends on Model, so you can remove this dependency.

service/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+    <parent>
+        <groupId>org.cbioportal.genome_nexus</groupId>
+        <artifactId>genome-nexus</artifactId>
+        <version>0.1.0</version>
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.cbioportal.genome_nexus</groupId>
+            <artifactId>model</artifactId>
```
ersinciftci: Persistence already depends on this, and this module depends on Persistence, so you can remove this dependency.

service/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
```
ersinciftci: Persistence already depends on this, and this module depends on Persistence, so you can remove this dependency.

util/pom.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ersinciftci: I think this module should merge with Service module.

web/pom.xml
```diff
@@ -34,6 +34,21 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.cbioportal.genome_nexus</groupId>
+      <artifactId>model</artifactId>
```
ersinciftci: Again an unnecessary dependency.

service/pom.xml
onursumer: Removing this dependency causes errors, seems like we still need this. It might be related to the `scope`.

ersinciftci: I forgot that test scoped dependencies are not transitive, but you can move it to the parent pom so every module can use it.

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/ExternalResourceTransformer.java
```diff
@@ -1,28 +0,0 @@
...
-import org.cbioportal.genome_nexus.annotation.util.Transformer;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Qualifier;
-import org.springframework.stereotype.Component;
-
-import java.io.IOException;
-import java.util.List;
-
-@Component
-public class ExternalResourceTransformer
```
onursumer: This file is also relocated

persistence/src/main/java/org/cbioportal/genome_nexus/util/Transformer.java
```diff
@@ -71,39 +69,4 @@
 
         return dbObjects;
     }
-
-    public static <T> List<T> mapJsonToInstance(String jsonString, Class<T> type) throws IOException
```
onursumer: Moved this method into `ExternalResourceTransformer`

persistence/src/main/java/org/cbioportal/genome_nexus/util/Transformer.java
```diff
@@ -71,39 +69,4 @@
...
-    /**
-     * Maps the given raw JSON string onto the provided class instance.
-     *
-     * @param jsonString    raw JSON string
-     * @param type          object class
-     * @param objectMapper  custom object mapper
-     * @return a list of instances of the provided class
-     * @throws IOException
-     */
-    public static <T> List<T> mapJsonToInstance(String jsonString, Class<T> type, ObjectMapper objectMapper) throws IOException
```
onursumer: Moved this method into `ExternalResourceTransformer`

service/src/main/java/org/cbioportal/genome_nexus/service/internal/ExternalResourceTransformer.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+@Component
+public class ExternalResourceTransformer
```
onursumer: This is not a new file, just relocated.

model/src/main/java/org/cbioportal/genome_nexus/model/EnsemblTranscript.java
inodb: @onursumer yeah const module sounds good

PR #76, opened 2017-10-24
Fallback on Ensembl API when DataAccessResourceFailureException encountered

angelicaochoa:
Fall back on using Ensembl REST service when Genome Nexus can't read from Mongo due to `DataAccessResourceFailureException`'s. 

Genome Nexus currently has a monitoring script for handling Mongo instability issues. This change will allow our annotations to remain uninterrupted while the monitoring script restarts the Mongo database.

### DIFF
--- a/annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/VEPVariantAnnotationService.java
+++ b/annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/VEPVariantAnnotationService.java
@@ -38,6 +38,7 @@ import org.cbioportal.genome_nexus.annotation.domain.*;
 import org.cbioportal.genome_nexus.annotation.service.*;
 
 import org.springframework.dao.DataIntegrityViolationException;
+import org.springframework.dao.DataAccessResourceFailureException;
 import org.springframework.stereotype.Service;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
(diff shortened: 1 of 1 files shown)


annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/VEPVariantAnnotationService.java
```diff
@@ -74,9 +75,17 @@ public VEPVariantAnnotationService(EnrichmentService enrichmentService,
 
     public VariantAnnotation getAnnotation(String variant)
     {
-        VariantAnnotation variantAnnotation = variantAnnotationRepository.findOne(variant);
+        boolean storeVariantJson = true;
```
onursumer: maybe name this variable `saveAnnotationJson` for consistency with the method name?

annotation/src/main/java/org/cbioportal/genome_nexus/annotation/service/internal/VEPVariantAnnotationService.java
```diff
@@ -93,15 +102,16 @@ public VariantAnnotation getAnnotation(String variant)
...
-                catch (DataIntegrityViolationException e) {
-                    // in case of data integrity violation exception, do not bloat the logs
-                    // this is thrown when the annotationJSON can't be stored by mongo
-                    // due to the variant annotation key being too large to index
-                    LOG.info(e.getLocalizedMessage());
+                if (storeVariantJson) {
+                    try {
+                        variantAnnotationRepository.saveAnnotationJson(variant, annotationJSON);
+                    }
+                    catch (DataIntegrityViolationException e) {
```
onursumer: Can we move this `catch` block outside, I mean as another `catch` statement for the outer `try`? By this way we can avoid a nested try statement. So that the code would simply look like this: 

```Java
if (saveAnnotationJson) {
    variantAnnotationRepository.saveAnnotationJson(variant, annotationJSON);
}
```

angelicaochoa:
@onursumer I've updated this pr following your recommendations. Thanks!

PR #83, opened 2017-11-03
Added PDB header service

onursumer:
Moved over the PDB info service implementation, with a simple key-value pair caching, from the https://github.com/genome-nexus/g2s/tree/master/pdb module, so that we can retire `pdb-annotation.war`.

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/PdbHeader.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/PdbHeader.java
@@ -1,0 +1,54 @@
+package org.cbioportal.genome_nexus.model;
+
+import java.util.Map;
+
+/**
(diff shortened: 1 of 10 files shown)


model/src/main/java/org/cbioportal/genome_nexus/model/PdbHeader.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * @author Selcuk Onur Sumer
+ */
+public class PdbHeader
+{
+    private String pdbId;
+    private String title;
+    private Map<String, Object> compound;
+    private Map<String, Object> source;
```
onursumer: Keeping this not-so-nice model as is for now in order to avoid front-end changes. We may want to revisit this model if we switch to a better web service in the future (XML or JSON).

PR #85, opened 2017-11-06
Rename get by hugo symbol to canonical-transcript/

inodb:
This makes more sense when looking at the other endpoints. The
/ensembl/transcript/hgnc should return all transcripts with that hugo symbol.
This moves all isoform override logic to a new endpoint. Also make ensembl
endpoint public, since we want to use it in cbioportal 1.10.0

Partially fix: https://github.com/cBioPortal/cbioportal/issues/3293

### DIFF
--- a/web/src/main/java/org/cbioportal/genome_nexus/web/EnsemblController.java
+++ b/web/src/main/java/org/cbioportal/genome_nexus/web/EnsemblController.java
@@ -5,7 +5,7 @@ import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import org.cbioportal.genome_nexus.model.EnsemblTranscript;
 import org.cbioportal.genome_nexus.service.EnsemblService;
-import org.cbioportal.genome_nexus.web.config.InternalApi;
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/cbioportal/genome_nexus/web/EnsemblController.java
```diff
@@ -129,7 +129,7 @@ public EnsemblController(EnsemblService ensemblService)
     }
 
     @ApiOperation(value = "Retrieves Ensembl transcripts by an Ensembl gene ID",
-        nickname = "fetchEnsemblTranscriptsByGeneIdGET")
+        nickname = "fetchCanonicalEnsemblTranscriptsByGeneIdGET")
```
onursumer: Also rename the method name to `fetchCanonicalEnsemblTranscriptsByGeneIdGET`?

inodb: good catch

inodb: @onursumer updated

PR #87, opened 2017-11-08
Improved error handling

onursumer:
- Added `NotFound` exceptions for the internal resources
- Added `NotFound` and `WebService` exceptions for external resources

### DIFF
--- a/web/src/main/java/org/cbioportal/genome_nexus/web/EnsemblController.java
+++ b/web/src/main/java/org/cbioportal/genome_nexus/web/EnsemblController.java
@@ -159,9 +159,11 @@ public class EnsemblController
             String isoformOverrideSource)
     {
         List<EnsemblTranscript> transcripts = new ArrayList<>();
+
         for (String hugoSymbol : hugoSymbols)
         {
-            transcripts.add(this.ensemblService.getCanonicalEnsemblTranscriptsByHugoSymbol(hugoSymbol, isoformOverrideSource));
+            transcripts.add(this.ensemblService.getCanonicalEnsemblTranscriptsByHugoSymbol(
+                hugoSymbol, isoformOverrideSource));
         }
 
         return transcripts;
--- a/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/PfamDomainRepositoryImpl.java
+++ b/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/PfamDomainRepositoryImpl.java
@@ -145,7 +145,6 @@ public class PfamDomainRepositoryImpl implements PfamDomainRepository
             this.pfamAList.addAll(this.parsePfamATSV());
 
             this.addNameAndDescription(pfamDomainList, pfamAList);
-            // TODO add color
 
             this.indexByTranscript = this.index(pfamDomainList, new TranscriptIndexKeyGetter());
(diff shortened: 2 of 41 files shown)


web/src/main/java/org/cbioportal/genome_nexus/web/EnsemblController.java
```diff
@@ -87,7 +92,11 @@ public EnsemblController(EnsemblService ensemblService)
 
         for (String proteinId : proteinIds)
         {
-            transcripts.addAll(this.ensemblService.getEnsemblTranscriptsByProteinId(proteinId));
+            try {
+                transcripts.addAll(this.ensemblService.getEnsemblTranscriptsByProteinId(proteinId));
+            } catch (EnsemblTranscriptNotFoundException e) {
+                e.printStackTrace();
```
ersinciftci: This doesn't seem right. You should probably have another service method that takes a list and returns a list, and wouldn't throw this exception.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/EnsemblServiceImpl.java
```diff
@@ -19,23 +20,55 @@
...
 
     @Override
-    public List<EnsemblTranscript> getEnsemblTranscriptsByTranscriptId(String transcriptId) {
-        return this.ensemblRepository.findByTranscriptId(transcriptId);
+    public List<EnsemblTranscript> getEnsemblTranscriptsByTranscriptId(String transcriptId)
+        throws EnsemblTranscriptNotFoundException
+    {
+        List<EnsemblTranscript> transcripts = this.ensemblRepository.findByTranscriptId(transcriptId);
+
+        if (transcripts.size() == 0) {
```
ersinciftci: You are probably using a path variable where you should be using a query variable, otherwise you wouldn't be returning 404 from a list-returning endpoint.

onursumer:
@inodb @ersinciftci Do you think it is a good idea to merge this PR as is and address the issues introduced with this PR while fixing #90?

service/src/main/java/org/cbioportal/genome_nexus/service/internal/EnsemblServiceImpl.java
inodb: @ersinciftci Thanks a ton for checking! This is a very good point and made clear that we actually had to change the endpoints quite drastically. Much appreciated!

inodb:
@onursumer yes agree, that we should tackle the endpoint query parameter in a separate PR.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/EnsemblServiceImpl.java
ersinciftci: @inodb No problem! Btw I realized I worded the comment poorly, list-returning endpoints can return 404 in some cases, like `/studies/{studyId}/samples` would return 404 if studyId is invalid, but it wouldn't have a `samples.size() == 0` check for it, it would have `study == null`, that's what I meant. @onursumer

onursumer:
@ersinciftci @inodb Reverted back certain `NotFound` exceptions. We may want to revisit those after we redefine the endpoints (#90)

Issue #89, opened 2017-11-10
Merge PFAM model with Ensembl model, and import the unified model into MongoDB


onursumer:
Current PFAM model is not very useful. Merge the PFAM model with Ensembl model and import a nested JSON structure into MongoDB.

onursumer:
I think we can still keep the pfam collection in the database, but model it only with these 3 fields:

```JavaScript
pfamDomain: {
    pfamAccession: "" // a unique id (reffered as pfamA_acc in pfamA.txt)
    name: "" // short name (referred as pfamA_id in pfamA.txt)
    description: "" // a little longer description
}
```

Simply importing the `pfamA.txt` file as a standalone collection would do the trick. 
 Let's move it from `src/main/resources/pfamA.txt` to `data`. Also, the column headers in that file might be confusing. `pfamA_acc` is the actual unique id, and `pfamA_id` is the short name (that we display in the domain rectangles in cBioPortal)

We can also keep the pfam endpoint, but it will be a simple endpoint where we retrieve domains only by id (accession).

And then we can model the `transcript` this way:

```JavaScript
transcript {
    transcriptId: "" // a unique id
    ...
    ...
    domains : [{
        pfamAccession: "" // refers to the pfam collection
        start: 10
        end: 20
    }, {
    ...
    }]
}
```

This way we will still need 2 API calls in cBioPortal to get the domain data, but I think that shouldn't be a huge deal. @inodb what do you think?

onursumer:
Fixed by #93 and #94

PR #93, opened 2017-11-13
Improved PFAM and Ensembl endpoints

### LINKED ISSUE #86: move cross-reference-controllor to ensembl/xref

### LINKED ISSUE #90: Refine pfam and ensembl endpoints
- Remodel Ensembl and PFAM, and remove pfam endpoint after fixing #89 
- Replace certain ensembl endpoint path variables with query parameters.

onursumer:
- Fixes #90 
- Also fixes #86 (previous endpoint is still accessible as an internal API, but marked as deprecated)

Improvements:
- Simplified and improved PFAM endpoint & model
- Improved ensemble endpoint
- Moved all the business logic from the controller layer to the service layer for all endpoints

We still need to:
- Update the Ensembl model (see #89)
- Update the importer (import scripts) for the PFAM file.


model/src/main/java/org/cbioportal/genome_nexus/model/PfamDomain.java
```diff
@@ -1,114 +1,44 @@
...
 
-import com.univocity.parsers.annotations.Parsed;
-import com.univocity.parsers.annotations.Trim;
+import org.springframework.data.mongodb.core.mapping.Document;
+import org.springframework.data.mongodb.core.mapping.Field;
 
 /**
  * @author Selcuk Onur Sumer
  */
+@Document(collection="pfam.domain")
```
onursumer: Instead of a reading a file at runtime, now we are using a mongo db collection for PFAM

model/src/main/java/org/cbioportal/genome_nexus/model/PfamA.java
```diff
@@ -1,46 +0,0 @@
-package org.cbioportal.genome_nexus.model;
-
-import com.univocity.parsers.annotations.Parsed;
-import com.univocity.parsers.annotations.Trim;
-
-/**
- * @author Selcuk Onur Sumer
- */
-public class PfamA
```
onursumer: This model is merged into `PfamDomain`

persistence/src/main/java/org/cbioportal/genome_nexus/persistence/PfamDomainRepository.java
```diff
@@ -1,17 +1,15 @@
...
 import org.cbioportal.genome_nexus.model.PfamDomain;
+import org.springframework.data.mongodb.repository.MongoRepository;
 
 import java.util.List;
 
 /**
  * @author Selcuk Onur Sumer
  */
-public interface PfamDomainRepository
+public interface PfamDomainRepository extends MongoRepository<PfamDomain, String>
```
onursumer: Now this is a proper `MongoRepository`

persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/PfamDomainRepositoryImpl.java
```diff
@@ -1,247 +0,0 @@
...
-import java.io.InputStreamReader;
-import java.io.Reader;
-import java.io.UnsupportedEncodingException;
-import java.util.*;
-
-/**
- * @author Selcuk Onur Sumer
- */
-@Component("defaultPfamDomainRepository")
-public class PfamDomainRepositoryImpl implements PfamDomainRepository
```
onursumer: We don't need a custom repository implementation anymore, we use `MongoRepository`

service/src/main/java/org/cbioportal/genome_nexus/service/MutationAssessorService.java
```diff
@@ -0,0 +1,23 @@
...
+import org.cbioportal.genome_nexus.model.MutationAssessor;
+import org.cbioportal.genome_nexus.model.VariantAnnotation;
+import org.cbioportal.genome_nexus.service.exception.MutationAssessorNotFoundException;
+import org.cbioportal.genome_nexus.service.exception.MutationAssessorWebServiceException;
+import org.cbioportal.genome_nexus.service.exception.VariantAnnotationNotFoundException;
+import org.cbioportal.genome_nexus.service.exception.VariantAnnotationWebServiceException;
+
+import java.util.List;
+
+public interface MutationAssessorService
```
onursumer: Added an interface for `MutationAssessorService`, previously we only had an implementation class

onursumer:
@inodb added mongodb import code for `pfamA.txt`, now pfam endpoint should work on heroku as well.

web/src/main/java/org/cbioportal/genome_nexus/GenomeNexusImporter.java
```diff
@@ -21,6 +21,7 @@
     final static String[][] IMPORT_RESOURCE_FILES = new String[][]{
         new String[]{"/ensembl_biomart_transcripts.txt", EnsemblRepositoryImpl.TRANSCRIPTS_COLLECTION},
         new String[]{"/ensembl_biomart_canonical_transcripts_per_hgnc.txt", EnsemblRepositoryImpl.CANONICAL_TRANSCRIPTS_COLLECTION},
+        new String[]{"/pfamA.txt", "pfam.domain"},
```
inodb: @onursumer why do you prefer here not to use constants?

onursumer: Actually, I don't prefer it :) I just couldn't find the right place to add this constant. We don't have a `PfamDomainRepositoryImpl`, I might have put the constant in the actual interface `PfamDomainRepository`. But then if I want to use the same constant in the model class, it would add a dependency to the `persistence` module in the `model` module, which we want to avoid. 

Same thing is also true for the constant `EnsemblRepositoryImpl.TRANSCRIPTS_COLLECTION`, we include it here but not in the model class. This is because our persistence annotations are still in the `model` module. Ideally we should find a way to add persistence annotations in the `persistence` module, not in the `model` module. But I wasn't able to find an easy way to do it, something similar to our `mixin`s in the web and service modules would be ideal.

inodb: ok, cool we can leave it for now

PR #94, opened 2017-11-14
Ensembl transcript nested pfam domains

inodb:
Only last commit is relevant (dependent on #93 ):

Changes the ensembl transcript model to include pfamDomains field:
```
{
    "geneId": "string",
    "pfamDomains": [
      {
        "pfamDomainEnd": 0,
        "pfamDomainId": "string",
        "pfamDomainStart": 0
      }
    ],
    "proteinId": "string",
    "proteinLength": 0,
    "transcriptId": "string"
  }
```

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/PfamDomainRange.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/PfamDomainRange.java
@@ -1,0 +1,39 @@
+package org.cbioportal.genome_nexus.model;
+
+import org.springframework.data.mongodb.core.mapping.Field;
+
+public class PfamDomainRange
+{
+    @Field(value="pfam_domain_id")
+    private String pfamDomainId;
+
+    @Field(value="pfam_domain_start")
+    private Integer pfamDomainStart;
+
+    @Field(value="pfam_domain_end")
+    private Integer pfamDomainEnd;
+
+    public String getPfamDomainId() {
+        return this.pfamDomainId;
+    }
+
+    public void setPfamDomainId(String pfamDomainId) {
+        this.pfamDomainId = pfamDomainId;
+    }
+
+    public Integer getPfamDomainStart() {
+        return this.pfamDomainStart;
+    }
+
+    public void setPfamDomainStart(Integer pfamDomainStart) {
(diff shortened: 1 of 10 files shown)


web/src/main/resources/application.properties.EXAMPLE
```diff
@@ -18,20 +18,12 @@ vep.isoform.overrides=uniprot:isoform_overrides_uniprot.txt,mskcc:isoform_overri
 # PDB header web service URL
 pdb.header_service_url=http://files.rcsb.org/header/PDB_ID.pdb
 
-# Ensembl biomart PFAM filename
-ensembl.biomart.pfam=ensembl_biomart_pfam.txt
```
onursumer: We may still want to keep these file names, and replace the hard-coded versions in the `GenomeNexusImporter`.

model/src/main/java/org/cbioportal/genome_nexus/model/PfamDomainRange.java
```diff
@@ -0,0 +1,39 @@
+package org.cbioportal.genome_nexus.model;
+
+import org.springframework.data.mongodb.core.mapping.Field;
+
+public class PfamDomainRange
```
onursumer: we should also add a mixin under `web` module for this model

web/src/main/resources/ensembl_biomart_transcripts.json
```diff
@@ -0,0 +1,5 @@
+{"transcript_stable_id":"ENST00000000233","gene_stable_id":"ENSG00000004059","protein_stable_id":"ENSP00000000233","protein_length":180.0,"domains":[{"pfam_domain_id":"PF00025","pfam_domain_start":7.0,"pfam_domain_end":176.0}]}
```
onursumer: Let's not forget to update this file with the full data :)

web/src/main/resources/application.properties.EXAMPLE
inodb: @onursumer I'm still not sure using `GenomeNexusImporter` is the right way to initialize the database. Making it configurable at this point seems like premature optimization

PR #95, opened 2017-11-15
Minor cleanup

onursumer:
- removed unused `PfamDomainMixin` fields
- removed unused static JS content


web/src/main/java/org/cbioportal/genome_nexus/web/mixin/PfamDomainMixin.java
```diff
@@ -14,11 +14,4 @@
 
     @ApiModelProperty(value = "PFAM domain description")
     private String description;
-
-    // TODO move this into Ensembl model...
-    @ApiModelProperty(value = "PFAM domain start")
-    private Integer pfamDomainStart;
```
onursumer: These are already added in `PfamDomainRangeMixin` with #94

web/src/main/resources/static/src/js/controller/GenomeNexus.js
```diff
@@ -62,17 +62,6 @@ function GenomeNexus(options)
...
-            },
-            download: {
-                content:[{
-                    links: [
-                        {href: 'href="files/hotspots.xls"',
```
onursumer: Trying to make it less obvious that we copied & pasted everything from cancer hotspots repo :)

PR #103, opened 2017-11-22
Add version controller and service

inodb:
Uses @project.version@ in new `maven.properties `file to pass the maven prop to spring. Then return on `/version` endpoint

Might be worth looking at actuator at some point: https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html

### DIFF
--- a/web/pom.xml
+++ b/web/pom.xml
@@ -85,6 +85,17 @@
           </execution>
         </executions>
       </plugin>
+      <!-- copy maven project properties to system variables -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>2.20.1</version>
+        <configuration>
+          <systemPropertyVariables>
+            <projectVersion>1.0</projectVersion>
+          </systemPropertyVariables>
+        </configuration>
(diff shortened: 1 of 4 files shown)


web/src/main/java/org/cbioportal/genome_nexus/web/InfoController.java
```diff
@@ -0,0 +1,27 @@
...
+@PublicApi
+public class InfoController
+{
+    @ApiOperation(value = "Retrieve Genome Nexus Version",
+        nickname = "fetchVersionGET")
+    @RequestMapping(value = "/version",
+        method = RequestMethod.GET,
+        produces = "application/json")
+    @ResponseBody
+    public Map<String,String> fetchVersionGET()
```
onursumer: As far as I remember, returning a Map was causing errors for swagger API client generator.

web/pom.xml
```diff
@@ -137,6 +148,7 @@
     <resources>
       <resource>
         <directory>src/main/resources</directory>
+        <filtering>true</filtering>
```
inodb: Necessary to replace `@` in properties file when it's copied over to `classes`

PR #116, opened 2017-12-15
Added cache support for Mutation Assessor data

onursumer:
Fix https://github.com/cbioportal/cbioportal/issues/3294

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/MutationAssessor.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/MutationAssessor.java
@@ -2,12 +2,15 @@ package org.cbioportal.genome_nexus.model;
 
 import org.springframework.data.annotation.Id;
 import org.springframework.data.mongodb.core.mapping.Document;
+import org.springframework.data.mongodb.core.mapping.Field;
 
-@Document(collection = "mutation_assessor")
+@Document(collection = "mutation_assessor.annotation")
 public class MutationAssessor
 {
     @Id
     private String input;
+
+    private String hgvs;
 
     private String hugoSymbol;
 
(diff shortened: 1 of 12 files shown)


model/src/main/java/org/cbioportal/genome_nexus/model/MutationAssessor.java
```diff
@@ -2,43 +2,46 @@
...
+    private String hgvs;
+
     private String hugoSymbol;
 
     private String referenceGenomeVariant;
     private String referenceGenomeVariantType;
 
     private String functionalImpact;
-    private double functionalImpactScore;
+    private Double functionalImpactScore;
```
onursumer: Using primitive types causes database mapping problems if the value is null. So, changed all `double`s to `Double`, and all `int`s to `Integer`.

model/src/main/java/org/cbioportal/genome_nexus/model/MutationAssessor.java
```diff
@@ -48,6 +51,15 @@ public void setInput(String input) {
         this.input = input;
     }
 
+    public String getHgvs() {
+        return hgvs;
```
onursumer: Previous code was overwriting the original `input` value returned by mutation assessor web service with the hgvs value. Now we are properly storing the hgvs value in a separate field.

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java
```diff
@@ -24,8 +24,16 @@
     protected ExternalResourceFetcher<T> fetcher;
     protected ResourceTransformer<T> transformer;
 
-    public BaseCachedExternalResourceFetcher(ResourceTransformer<T> transformer)
+    public BaseCachedExternalResourceFetcher(String collection,
+                                             R repository,
+                                             Class<T> type,
+                                             ExternalResourceFetcher<T> fetcher,
+                                             ResourceTransformer<T> transformer)
```
onursumer: Updated the constructor in order to force the child classes to provide these 5 arguments.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/MutationAssessorServiceImpl.java
```diff
@@ -71,26 +77,27 @@ public MutationAssessor getMutationAssessor(VariantAnnotation annotation)
...
-            List<MutationAssessor> list = this.externalResourceFetcher.fetchInstances(variant);
-
-            if (list.size() != 0) {
-                mutationAssessorObj = list.get(0);
-            }
+            // get the annotation from the web service and save it to the DB
+            mutationAssessor = cachedExternalResourceFetcher.fetchAndCache(variant);
 
-            if (mutationAssessorObj != null) {
-                mutationAssessorObj.setInput(hgvs);
```
onursumer: This is where we were overwriting the actual input value with hgvs notation. This is not a good practice :)

PR #122, opened 2017-12-28
Made findOneByHugoSymbol query case insensitive for EnsemblRepository

onursumer:
Partial fix for https://github.com/cBioPortal/cbioportal/issues/3644

### DIFF
--- a/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/EnsemblRepositoryCustom.java
+++ b/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/EnsemblRepositoryCustom.java
@@ -35,5 +35,5 @@ import org.cbioportal.genome_nexus.model.EnsemblTranscript;
 
 public interface EnsemblRepositoryCustom
 {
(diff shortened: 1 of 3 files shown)


jjgao:
Should we normalize the data in the database?

onursumer:
I think we should update our import scripts and rerun them for all the files we have.

Ideally, we should have a proper importer module for Genome Nexus. Currently we have separate python scripts doing the job.

onursumer:
@jjgao: The number of genes in the portal database is 145084, whereas number of genes in the genome nexus DB with valid transcript mapping is 33670.

Also, in portal database I see gene names like `ZZZ3_PS113`, `ZZZ3_PS130`, `ZZZ3_PS131`, `ZZZ3_PS135`, `ZZZ3_PS305`, etc. But in the genome nexus database we only have `ZZZ3`.

sheridancbio:
I've approved, but I wonder about why we need to be sloppy about case.  Luckily, within a species there does not seem to be ambiguity when you ignore case in genes symbols.

onursumer:
Thanks @sheridancbio! 

Seems like the query coming from cBioPortal side is always uppercase. Genes like `C3orf70` fail when queried as `C3ORF70`. With this PR now we can support case insensitive queries.

PR #131, opened 2018-01-18
Genomenexus doc improvement

pieterlukasse:
- Added mongoDB step to documentation and also 
- Added info on how to let it run if mongoDB is running in docker


inodb:
Thanks @pieterlukasse that looks great! I see rc misses a few commits from master, that's prolly why these other commits are showing up.

PR #132, opened 2018-01-24
Add Hugo Symbols to transcript endpoint

inodb:
Fix #128

PR #134, opened 2018-01-30
Taking variant classification into account when filtering hotspots

onursumer:
Improved cancer hotspot service so that we better filter hotspots for a given transcript.

Needed to move over and refactor some code from `genome-nexus-annotation-pipeline`. Eventually we should move everything there to genome nexus.

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/annotation/CanonicalTranscriptResolver.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/annotation/CanonicalTranscriptResolver.java
@@ -1,0 +1,51 @@
+package org.cbioportal.genome_nexus.service.annotation;
+
+import org.cbioportal.genome_nexus.model.TranscriptConsequence;
+import org.cbioportal.genome_nexus.model.VariantAnnotation;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Component
+public class CanonicalTranscriptResolver
+{
+    private final TranscriptConsequencePrioritizer consequencePrioritizer;
+
+    @Autowired
+    public CanonicalTranscriptResolver(TranscriptConsequencePrioritizer consequencePrioritizer)
+    {
+        this.consequencePrioritizer = consequencePrioritizer;
+    }
+
+    public TranscriptConsequence resolve(VariantAnnotation variantAnnotation)
+    {
+        List<TranscriptConsequence> transcripts = new ArrayList<>();
+
+        for (TranscriptConsequence transcript : variantAnnotation.getTranscriptConsequences())
+        {
+            if (transcript.getTranscriptId() != null &&
+                transcript.getCanonical() != null &&
+                transcript.getCanonical().equals("1"))
+            {
+                transcripts.add(transcript);
+            }
+        }
+
+        // only one transcript marked as canonical
+        if (transcripts.size() == 1) {
+            return transcripts.iterator().next();
+        }
+        else if (transcripts.size() > 1) {
+            return this.consequencePrioritizer.transcriptWithMostSevereConsequence(
+                transcripts, variantAnnotation.getMostSevereConsequence());
+        }
+        // no transcript marked as canonical (list.size() == 0),
+        // use most severe consequence to decide which one to pick among all available
+        else {
+            return this.consequencePrioritizer.transcriptWithMostSevereConsequence(
+                variantAnnotation.getTranscriptConsequences(), variantAnnotation.getMostSevereConsequence());
+        }
+    }
+}
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/annotation/VariantClassificationResolver.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/annotation/VariantClassificationResolver.java
@@ -1,0 +1,200 @@
+package org.cbioportal.genome_nexus.service.annotation;
+
+import org.cbioportal.genome_nexus.model.TranscriptConsequence;
+import org.cbioportal.genome_nexus.model.VariantAnnotation;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+@Component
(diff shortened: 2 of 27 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/annotation/VariantClassificationResolver.java
```diff
@@ -0,0 +1,163 @@
...
+    {
+        String variantClassification = null;
+        String variantType = this.variantTypeResolver.resolve(variantAnnotation);
+        Boolean isInframe = this.isInframe(variantAnnotation);
+
+        // if not null, try to resolve the variant classification for the provided transcript
+        if (transcriptConsequence != null)
+        {
+            variantClassification = this.resolveVariantClassification(
+                this.consequencePrioritizer.pickHighestPriorityConsequence(transcriptConsequence.getConsequenceTerms()),
```
onursumer: @jjgao @inodb this logic is adopted from the annotation pipeline repository, and it only resolves to a single mutation type. I think this is what we want when annotating MAF files. But for hotspot filtering purposes should we consider all consequence terms and return a list of mutation types instead of just one result?

service/src/main/java/org/cbioportal/genome_nexus/service/annotation/CanonicalTranscriptResolver.java
```diff
@@ -0,0 +1,51 @@
...
+    public CanonicalTranscriptResolver(TranscriptConsequencePrioritizer consequencePrioritizer)
+    {
+        this.consequencePrioritizer = consequencePrioritizer;
+    }
+
+    public TranscriptConsequence resolve(VariantAnnotation variantAnnotation)
+    {
+        List<TranscriptConsequence> transcripts = new ArrayList<>();
+
+        for (TranscriptConsequence transcript : variantAnnotation.getTranscriptConsequences())
```
inodb: I guess this is copied over from the importer but what is canonical differs for mskcc/uniprot/ensembl. Now we are always using canonical ensembl transcript,  if that isn't found the most severe one. Doesn't seem what we want

service/src/main/java/org/cbioportal/genome_nexus/service/annotation/VariantClassificationResolver.java
inodb: Yeah after merging #132 we should be able to match on hugo symbol and consequence for any of the transcripts. If one of them matches it's a hotspot

service/src/main/java/org/cbioportal/genome_nexus/service/annotation/CanonicalTranscriptResolver.java
onursumer: That's right, copied over from importer. With the latest changes to this PR, we are not relying on the canonical transcript anymore when figuring out the hotspots. We may revisit the `CanonicalTranscriptResolver` when we need to use it...

onursumer:
Added tests for the hotspot mutation filtering logic. Ideally, we should also test the stuff that we moved over from annotation pipeline. That part is currently not in use for the cancer hotspots endpoint though.

web/src/test/java/org/cbioportal/genome_nexus/web/CancerHotspotsIntegrationTest.java
```diff
@@ -0,0 +1,191 @@
...
+        String baseUrl = "http://localhost:38888/cancer_hotspots/genomic/";
+
+        // genomic location strings (needed for GET requests)
+        String[] genomicLocations = {
+            "7,140453193,140453193,T,C", // recurrent and 3D
+            "17,37880220,37880220,T,C", // recurrent
+            "17,37879794,37879794,G,C", // recurrent
+            "12,25398284,25398285,CC,GA", // recurrent and 3D
+            "12,25380271,25380271,C,T", // 3D
+            "10,89692940,89692940,C,T" // none
```
inodb: @onursumer should we handle `chr1` as well?

inodb: things prefixed with `chr*` basically

inodb:
🥞 🌮

web/src/test/java/org/cbioportal/genome_nexus/web/CancerHotspotsIntegrationTest.java
inodb: @onursumer what about indels? Splice? Would be good to just indicate those aren't handled yet (if they're not)

onursumer: good point, probably we are not handling the `chr*` case right now.

inodb: yeah not a big deal, but would be good to include. Also for indels the format is different for vcf vs maf. So good to have some parameter to indicate which one we are using

onursumer: @inodb indels are handled, see the `testIndelHotspots` method below :) 

Splice is probably not gonna work though

inodb: @onursumer oops missed those

web/src/test/java/org/cbioportal/genome_nexus/web/CancerHotspotsIntegrationTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+    @Test
+    public void testIndelHotspots()
+    {
+        String baseUrl = "http://localhost:38888/cancer_hotspots/genomic/";
+
+        // genomic location strings (needed for GET requests)
+        String[] genomicLocations = {
+            "7,55242466,55242480,GAATTAAGAGAAGCA,-",
+            "7,55249002,55249003,-,CAGCGTGGA"
```
inodb: @onursumer this is great! Looks like we are using MAF format here. Maybe good to have some parameter indicating that. 

In the same vain - how are impossible ref/alt/position combos handled?

inodb: ill just filed an issue for my comments so we can merge this.

https://github.com/genome-nexus/genome-nexus/issues/140

PR #136, opened 2018-02-03
Move data/ to genome-nexus-importer

inodb:
Move data transformation and import to a different repo. For development one
can use the docker image from that repo. Add a docker compose file that uses
that approach. Change documentation accordingly.

See also https://hub.docker.com/r/genomenexus/gn-mongo/ and the repo that is based on: https://github.com/genome-nexus/genome-nexus-importer


README.md
```diff
@@ -23,47 +23,34 @@ programmatic interface
...
 ```
-spring.data.mongodb.uri=mongodb://gn-mongo:27017/annotator
+# change latest to different version if necessary (only need to run this once)
+docker run --name=gn-mongo --restart=always -p 27017:27017 -d genomenexus/gn-mongo:latest 
+mvn  -DskipTests clean install
+java -jar web/target/web-*.war
 ```
+### Alternative 3 - install mongoDB locally and run with local java
+Install mongoDB manually. Then follow instructions in
+[genome-nexus-importer](https://github.com/genome-nexus/genome-nexus-importer)
```
pieterlukasse: following these instructions applies to all alternatives, right?

inodb: @pieterlukasse actually it does not, because when you use docker compose, you don't have to initialize the database. It just pulls  mongo with the data imported

inodb: Maybe I have to clarify in docs

inodb: clarified

inodb:
Got verbal approval from @onursumer

PR #133, opened 2018-01-29
Add exon information to transcript IDs in Ensembl Controller

dionnezaal:
# What?
This PR adds exon information to the current `EnsemblTranscript` model, called `ExonRange`:
 ```
ExonRange {
exonId (string): Exon id,
exonStart (integer): Start position of exon,
exonEnd (integer): End position of exon ,
rank (integer): Number of exon in transcript ,
strand (integer): Strand exon is on, -1 for - and 1 for +,
version (integer): Exon version
}
```
# Why?
See issue #84 

## This PR includes:
- Added model `ExonRange` to `EnsemblTranscript` Model
- Documentation and transformation script to retrieve exon information from Ensembl and add it to the current transcript information file
- Replacement of script that adds domains to transcript information file (`add_nested_domain_to_ensembl_transcript.py`) with a script that adds both domain and exon information to the transcript information file (`add_domains_and_exon_information_to_ensembl_transcript.py`). The previous script was replaced due to time it took to run. Replacement script can add the information in under a minute

PR #139, opened 2018-02-15
Add exon information to Ensembl transcripts

inodb:
Add exon information to the current EnsemblTranscript model, called
ExonRange:

```
ExonRange {
    exonId (string): Exon id,
    exonStart (integer): Start position of exon,
    exonEnd (integer): End position of exon ,
    rank (integer): Number of exon in transcript ,
    strand (integer): Strand exon is on, -1 for - and 1 for +,
    version (integer): Exon version
}
```

- Add model ExonRange to EnsemblTranscript Model

Original from @dionnezaal , see PR #133


docker-compose.yml
```diff
@@ -12,7 +12,7 @@ services:
     depends_on:
       - db
   db:
-    image: genomenexus/gn-mongo:latest
+    build: https://github.com/inodb/genome-nexus-importer.git#add-exon-info-to-transcript
```
pieterlukasse: What does this do? Does it just make an image using the Dockerfile here https://github.com/inodb/genome-nexus-importer.git#add-exon-info-to-transcript? Should this be pointing to `github.com/inodb/` ?

pieterlukasse:
@inodb looks good. Just some minor questions above.

docker-compose.yml
dionnezaal: I think necessary because PR for exon information is not yet merged. When merged this should be changed to `github.com/genome-nexus/genome-nexus-importer.git` I guess.

inodb: @pieterlukasse yes it makes it using the Dockerfile there. You can even select different dockerfiles in the same repo if required. We should indeed change it to point to main fork once merged as @dionnezaal mentioned

PR #147, opened 2018-03-08
Added protein position resolver

onursumer:
- In case of `dup` mutations, protein position is derived from protein change value when filtering hotspots.
- `ProteinChangeResolver` is mainly moved over from `genome-nexus-annotation-pipeline`. Ideally we should add some tests for that class too.


inodb:
@onursumer we should prolly add these functions as a dependency in genome-nexus-annotation-pipeline as well

onursumer:
@inodb Ideally, we should completely refactor https://github.com/genome-nexus/genome-nexus-annotation-pipeline/blob/master/annotator/src/main/java/org/cbioportal/annotator/internal/GenomeNexusImpl.java and move most of the logic into Genome Nexus.

PR #149, opened 2018-03-14
Added batch fetch and cache support for VEP annotation fetcher

onursumer:
Do not merge until https://github.com/Ensembl/ensembl-vep/issues/156 is fixed


persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/BaseGenericMongoRepository.java
```diff
@@ -0,0 +1,27 @@
...
+    public BaseGenericMongoRepository(MongoTemplate mongoTemplate)
+    {
+        this.mongoTemplate = mongoTemplate;
+    }
+
+    @Override
+    public void saveDBObjects(String collection, List<DBObject> dbObjects)
+    {
+        // TODO use bulk save instead!
+        for (DBObject dbObject: dbObjects) {
```
onursumer: Saving to DB is still done one by one, need to replace with a bulk save

web/src/main/resources/application.properties.EXAMPLE
```diff
@@ -1,6 +1,7 @@
 # VEP web API URL with additional parameters. Note that "VARIANT" field will be replaced with
 # the actual variant value before each VEP request.
 vep.url=http://grch37.rest.ensembl.org/vep/human/hgvs/VARIANT?content-type=application/json&xref_refseq=1&ccds=1&canonical=1&domains=1&hgvs=1&numbers=1&protein=1
+vep.max_page_size=300
```
onursumer: This is the max page size allowed by public VEP service, it might be possible to increase this for a private instance.

inodb: maybe add comment here in the code itself

service/src/main/java/org/cbioportal/genome_nexus/service/cached/CachedVariantAnnotationFetcher.java
```diff
@@ -1,25 +1,97 @@
...
 public class CachedVariantAnnotationFetcher extends BaseCachedExternalResourceFetcher<VariantAnnotation, VariantAnnotationRepository>
 {
+    public static final String FIXED_ENTRY = "AGT:c.803T>C";
+
     @Autowired
     public CachedVariantAnnotationFetcher(ExternalResourceTransformer<VariantAnnotation> transformer,
                                           VariantAnnotationRepository repository,
-                                          VEPDataFetcher fetcher)
+                                          VEPDataFetcher fetcher,
+                                          @Value("${vep.max_page_size}") Integer maxPageSize)
```
inodb: @onursumer good to specify default here as well `${vep.max_page_size:300}`. This way it doesn't crash when u don't update your properties file

PR #153, opened 2018-03-21
Added null check to prevent NPE when identifying dup protein changes

onursumer:
Locally tested with failing cBioPortal queries (e.g: http://www.cbioportal.org/index.do?session_id=5ab2808e498eb8b3d5654fc7), missing hotspots show up again with this fix.


inodb:
@onursumer this feels like something we could catch with typing. Maybe there's a way to define which fields might be null? Thoughts?

PR #160, opened 2018-03-28
Moved over changes from genome nexus annotation pipeline

onursumer:
Moved over the related change from https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/61


onursumer:
@inodb We cannot directly move over the tests since it relies on `MutationRecord` model, which we don't have in Genome Nexus. With some effort we can add similar tests though. And, we should definitely add tests for all our "resolver" classes, not just for this one.

inodb:
👌

PR #172, opened 2018-04-13
Moved over remaining annotation resolvers

onursumer:
- Moved over remaining annotation resolvers from https://github.com/genome-nexus/genome-nexus-annotation-pipeline.
- Added a new service and endpoints to get annotation summary by genomic location.
- Couple of bug fixes and improvements.

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/annotation/EntrezGeneIdResolver.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/annotation/EntrezGeneIdResolver.java
@@ -8,10 +8,6 @@ import org.cbioportal.genome_nexus.service.exception.EnsemblWebServiceException;
 import org.jetbrains.annotations.Nullable;
 import org.springframework.stereotype.Component;
 
-import java.util.List;
-import java.util.Optional;
-import java.util.stream.Collectors;
-
 @Component
 public class EntrezGeneIdResolver
 {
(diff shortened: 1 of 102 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/annotation/EntrezGeneIdResolver.java
```diff
@@ -0,0 +1,82 @@
...
+        String entrezGeneId = null;
+
+        if (transcriptConsequence != null &&
+            transcriptConsequence.getGeneId() != null &&
+            !transcriptConsequence.getGeneId().trim().isEmpty())
+        {
+            // TODO in memory cache?
+            // check hashmap first before hitting api for gene cross-ref data
+//            if (ensemblAccessionEntrezIdMap.containsKey(transcriptConsequence.getGeneId())) {
+//                return ensemblAccessionEntrezIdMap.get(transcriptConsequence.getGeneId());
```
onursumer: This is the original code which uses an in-memory cache. Maybe we should cache the xref API call results in our database instead.

service/src/main/java/org/cbioportal/genome_nexus/service/annotation/ProteinChangeResolver.java
```diff
@@ -91,11 +91,15 @@ else if (hgvsc != null) {
...
+        if (hgvsp == null) {
+            return null;
+        }
+
+        String hgvspShort = hgvsp;
 
         for (int i = 0; i < 24; i++) {
             if (hgvsp.contains(AA3TO1[i][0])) {
-                hgvspShort = hgvsp.replaceAll(AA3TO1[i][0], AA3TO1[i][1]);
+                hgvspShort = hgvspShort.replaceAll(AA3TO1[i][0], AA3TO1[i][1]);
```
onursumer: This is a bug fix. And actually the bug is already in production, but not critical at this point.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VEPEnrichmentService.java
```diff
@@ -68,7 +68,8 @@ public void registerEnricher(String id, AnnotationEnricher enricher)
         // initiate enricher list if not initiated yet
         if (enrichers == null)
         {
-            enrichers = new HashMap<>();
+            // linked hash map ensures that enrichers are executed by the same order they are registered
+            enrichers = new LinkedHashMap<>();
```
onursumer: This is important to make sure that certain enrichers are always executed first. For example, we want the isoform override enricher to run first so that we pick the correct canonical transcript.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationServiceImpl.java
```diff
@@ -200,12 +249,19 @@ private EnrichmentService initPostEnrichmentService(String isoformOverrideSource
...
             postEnrichmentService.registerEnricher("cancerHotspots", enricher);
         }
+
         if (fields != null && fields.contains("mutation_assessor"))
         {
             AnnotationEnricher enricher = new MutationAssessorAnnotationEnricher(mutationAssessorService);
             postEnrichmentService.registerEnricher("mutation_assessor", enricher);
         }
 
+        if (fields != null && fields.contains("annotation_summary"))
```
onursumer: Annotation summary (which mainly contains all the fields that `genome-nexus-annotation-pipeline` needs) as an optional field. There is also a separate endpoint where one can get just the summary.

PR #178, opened 2018-05-07
Add simple rule to ignore invalid hgvs for vep

inodb:


service/src/main/java/org/cbioportal/genome_nexus/service/cached/CachedVariantAnnotationFetcher.java
```diff
@@ -36,6 +36,11 @@ public CachedVariantAnnotationFetcher(ExternalResourceTransformer<VariantAnnotat
             maxPageSize);
     }
 
+    @Override
+    protected Boolean isValidId(String id) {
+        return !id.contains("N") && !id.contains("-") && !id.contains("undefined") && id.split(":")[1].matches(".*\\d+.*");
```
onursumer: `id.split(":")[1]` may cause a problem if the id doesn't contain `:`. Maybe also add the condition `id.contains(":")`

inodb: good catch

PR #179, opened 2018-05-07
Fix and improvements for the batch fetch and cache logic

onursumer:
- Now adding fixed entry to the beginning of each and every chunk for the batch query
- Added check for invalid ids for batch query as well


service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java
```diff
@@ -133,6 +134,9 @@ public T fetchAndCache(String id) throws ResourceMappingException
         // remove already cached ids from the set, so that we don't query again
         needToFetch.removeAll(alreadyCached);
 
+        // also remove invalid ids
```
onursumer: @inodb seems like we missed this one in #178

PR #181, opened 2018-05-10
Cancer Hotspots endpoint improvements

onursumer:
- Completely decoupled Genome Nexus from Cancer Hotspots Web API
- Using database instead of in-memory cache for cancer hotspots 
- Now we filter and import hotspots data with the importer module, related PR (already merged): genome-nexus/genome-nexus-importer/pull/7


service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
```diff
@@ -55,97 +51,50 @@
 @Service
 public class CancerHotspotServiceImpl implements CancerHotspotService
 {
-    private HotspotCache cache;
-
-    private final CancerHotspotDataFetcher cancerHotspotDataFetcher;
-    private final CancerHotspot3dDataFetcher cancerHotspot3dDataFetcher;
```
onursumer: removed the cache & the fetchers, using DB instead

service/src/main/java/org/cbioportal/genome_nexus/service/internal/HotspotFilter.java
```diff
@@ -15,6 +15,9 @@
 @Component
 public class HotspotFilter
 {
+    public static Double MISSENSE_FRACTION_THRESHOLD = 0.25;
+    public static Integer MISSENSE_COUNT_THRESHOLD = 1;
```
onursumer: @inodb should we make this configurable via `application.properties`?

service/src/main/java/org/cbioportal/genome_nexus/service/mixin/HotspotMixin.java
```diff
@@ -1,9 +0,0 @@
-package org.cbioportal.genome_nexus.service.mixin;
-
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-
-@JsonIgnoreProperties(ignoreUnknown = true)
-public class HotspotMixin
```
onursumer: We don't need a Mixin in the service layer anymore

web/src/test/java/org/cbioportal/genome_nexus/web/CancerHotspotsIntegrationTest.java
```diff
@@ -22,18 +27,40 @@
...
 public class CancerHotspotsIntegrationTest
 {
+    private static List<Hotspot> HOTSPOTS;
+
+    @BeforeClass
+    public static void readMockFile() throws IOException
+    {
+        JsonToObjectMapper objectMapper = new JsonToObjectMapper();
+        // reading the entire DB dump for a full scale test
+        HOTSPOTS = objectMapper.readHotspots("hotspots_db_dump.json");
```
onursumer: Reading a `json` dump into memory first as `Hotspot` instances, and then inserting into the DB. There might be a faster, more straightforward way to do this.

web/src/test/java/org/cbioportal/genome_nexus/web/CancerHotspotsIntegrationTest.java
```diff
@@ -22,18 +27,40 @@
...
+    @Autowired
+    private HotspotRepository hotspotRepository;
+
+    @Before
+    public void setupDB()
+    {
+        // only init once
+        if (this.hotspotRepository.count() == 0) {
+            // save the entire dump into the embedded mongoDB
+            this.hotspotRepository.insert(HOTSPOTS);
```
onursumer: Ideally we should do this in a `@BeforeClass` method, but the repository is not yet autowired in that case, so this is a sort of workaround.

web/src/test/java/org/cbioportal/genome_nexus/web/mock/HotspotMixin.java
```diff
@@ -0,0 +1,32 @@
+package org.cbioportal.genome_nexus.web.mock;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class HotspotMixin
```
onursumer: This mixin is used only for test purposes, while reading from a json DB dump. There might be a way to bypass this.

web/src/test/java/org/cbioportal/genome_nexus/web/CancerHotspotsIntegrationTest.java
inodb: Yeah also we would have to keep the dump in sync. It's most comprehensible I think to only import the relevant data for testing, so it's easy to see what kind of data is expected. You can use e.g. `jq` to get the relevant data.

Or if necessary to have the full dump, do a complete integration and use the actual data that we would import instead of having to sync this file.

onursumer: @inodb At this point I don't think we need the full dump. For now let's include only the hotspots related to the test then.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/HotspotFilter.java
onursumer: Makes sense if in Genome Nexus we are only interested in the subset of hotspots mutations where `missense_fraction >= 25% && missense_count > 1`. I am fine with that solution if it is not a problem to not import all the hotspots mutations we have in Cancer Hotspots website.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/HotspotFilter.java
```diff
@@ -64,6 +65,9 @@ public Boolean typeMatches(String hotspotType, String variantClassification)
         if (hotspotType.contains("single residue")) {
             typeMatches = variantClassification.toLowerCase().contains("missense");
         }
+        else if (hotspotType.contains("splice site")) {
```
inodb: nice

PR #192, opened 2018-06-07
Removed the leading p. from the hgvsp short string

onursumer:
Further shortened the `hgvspShort` value by removing the leading `p.` string.


angelicaochoa:
@onursumer I confirmed that this change shouldn't affect the annotation pipeline either. lgtm

PR #195, opened 2018-06-15
Add UTRs to /ensembl/transcript

sandertan:
# What?
Add untranslated regions (UTRs) to /ensembl/transcript info which makes it possible to infer the coding region of a transcript by:
(length all exons) - (length all UTRs) = (length coding sequence)

# Why?
The coding sequence can be mapped to PFAM sequence, which is useful for mapping PFAM domains. PFAM domains will be mapped to coding sequence in an upcoming feature for fusion visualization.

# Notes
This PR is dependent on the updated importer and database, PR in https://github.com/genome-nexus/genome-nexus-importer/pull/10


web/src/main/java/org/cbioportal/genome_nexus/web/mixin/EnsemblTranscriptMixin.java
```diff
@@ -25,7 +25,9 @@
     @ApiModelProperty(value = "Pfam domains", position=5, dataType="List")
     private List<PfamDomainRangeMixin> pfamDomains;
 
-    @ApiModelProperty(value = "Exon information", position=6, dataType="List")
-    private List<ExonRangeMixin> exons;
+    @ApiModelProperty(value = "Exon information", position=7, dataType="List")
```
rnugraha: maybe minor but why skipping position 6?

sandertan: Full code:

```
    @ApiModelProperty(value = "Hugo symbols", position=6, dataType="List")
    private List<String> hugoSymbols;

    @ApiModelProperty(value = "Pfam domains", position=5, dataType="List")
    private List<PfamDomainRangeMixin> pfamDomains;

    @ApiModelProperty(value = "Exon information", position=7, dataType="List")
    private List<ExonMixin> exons;

    @ApiModelProperty(value = "UTR information", position=8, dataType="List")
    private List<UntranslatedRegionMixin> utrs;
```

Position 6 was already taken.

jjgao:
Why do we have PFAM data in this endpoint? 

I think PFAM data should be served as a separate end point. We plan to support all kinds of annotation (PFAM is one of them). 

`/ensembl/transcript` does not seem like a proper endpoint to serve annotation data beyong basic information about the transcript.

Thoughts?

pieterlukasse:
@jjgao just discussed this with @sandertan and @inodb now. We think it could be nice to support a detailed projection level that does include the pfam data (similar to how cBioPortal api works). Maybe @inodb can add 2 projection levels for the transcript endpoint (SUMMARY, DETAILED)?

inodb:
@pieterlukasse opened issue for the multiple projections: https://github.com/genome-nexus/genome-nexus/issues/197

docker-compose.yml
```diff
@@ -12,7 +12,7 @@ services:
     depends_on:
       - db
   db:
-    build: https://github.com/genome-nexus/genome-nexus-importer.git#b056620a52368edb1c3baa6cec47cd5d3167a782
+    build: https://github.com/genome-nexus/genome-nexus-importer.git#83ae860fd388b51198ecc73e2f572a3b7e1f6a07
```
inodb: @sandertan this doesn't work i believe, can only refer to branches e.g. https://github.com/thehyve/genome-nexus-importer.git#add_UTRs

sandertan: @inodb Hmm I think it works in my instance after a:
```
mvn  -DskipTests clean install
docker compose up --build -d
```

screen shot 2018-06-21 at 13 56 50

jjgao:
@pieterlukasse @inodb @sandertan I am not convinced of using detailed projection in this case.

Detailed projection would make sense when there is a dependency on the detailed projection, e.g. a sample depends on (belong to) a patient which depends on a study. In that case, you can add multi-level detailed projection for more detailed information about patient and even study.

But in this case, a transcript does not depend on pfam domains. It is actually the other way around.

I think we should resist the temptation of lumping heterogenous data into the same API (that does not naturally host the data) just to meet one use case. It's important to keep genome nexus api general.

With that said, I am ok with the idea of have a segment-annotation endpoint that aggregate different types of annotations on a protein segment, e.g. pfam, secondary structures, exon (?).

inodb:
@jjgao those are all good points - i updated to issue to reflect that. Let's merge this n eventually move this over to some sort of segment-annotation endpoint

jjgao:
@inodb thanks, but which issue did you update?

inodb:
@jjgao https://github.com/genome-nexus/genome-nexus/issues/197

PR #204, opened 2018-07-11
Fix #200 convert 23/24-X/Y

### LINKED ISSUE #200: Convert X/Y to corresponding chromosome before annotation

inodb:
- Add methods to NotationConverter class to normalize chr23 to X, and chr24 to
  Y. Taking care of both using variants prefixed with chr and the 23/x and 24/y
  issue.
- Add tests for the variant annotation service. The NotionConverter utility
  functions are v simple, so seems better to test the annotation service
  directly.


service/src/test/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationServiceTest.java
```diff
@@ -42,6 +44,9 @@
     @Mock
     private CancerHotspotServiceImpl cancerHotspotService;
 
+    @Spy
+    private NotationConverter notationConverter;
```
inodb: using spy decorator to include actual `@Component` utility class, thanks @onursumer

PR #210, opened 2018-07-31
Use _splice hgvspShort when splice is most severe

inodb:
Fix https://github.com/cBioPortal/cbioportal/issues/2236

When there are multiple consequence terms and splice is the most severe, use *_splice for hgvspShort.

```
{
"transcriptId": "ENST00000275493",
"codonChange": "ctG/ctT",
"entrezGeneId": "1956",
"consequenceTerms": "splice_region_variant,synonymous_variant",
"hugoGeneSymbol": "EGFR",
"hgvspShort": "L210=",
"hgvsp": "p.Leu210=",
"hgvsc": "ENST00000275493.2:c.630G>T",
"proteinPosition": {
"start": 210,
"end": 210
},
"refSeq": "NM_005228.3",
"variantClassification": "Splice_Region"
},
```
becomes
```
{
"transcriptId": "ENST00000275493",
"codonChange": "ctG/ctT",
"entrezGeneId": "1956",
"consequenceTerms": "splice_region_variant,synonymous_variant",
"hugoGeneSymbol": "EGFR",
"hgvspShort": "X210_splice",
"hgvsc": "ENST00000275493.2:c.630G>T",
"proteinPosition": {
"start": 210,
"end": 210
},
"refSeq": "NM_005228.3",
"variantClassification": "Splice_Region"
}
```

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/annotation/VariantClassificationResolver.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/annotation/VariantClassificationResolver.java
@@ -132,7 +132,7 @@ public class VariantClassificationResolver
     {
         String defaultValue = "Targeted_Region";
 
-        if (variant == null || variantType == null) {
+        if (variant == null || (variantType == null && !variant.contains("splice"))) {
             return defaultValue;
         }
         else {
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/annotation/ProteinChangeResolver.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/annotation/ProteinChangeResolver.java
@@ -155,6 +155,8 @@ public class ProteinChangeResolver
         try {
             String[] aaParts = transcriptConsequence.getAminoAcids().split("/");
 
+            // TODO: always using get(0) here, maybe better to get most
+            // impactful consequence and use that?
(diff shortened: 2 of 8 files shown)


inodb:
Once this is merged we should add `7   55220240    55220240    G   T` to `test/data/minimal_example.in.txt` of the annotation pipeline

service/src/main/java/org/cbioportal/genome_nexus/service/annotation/VariantClassificationResolver.java
```diff
@@ -132,7 +132,7 @@ private String resolveVariantClassification(String variant, String variantType,
     {
         String defaultValue = "Targeted_Region";
 
-        if (variant == null || variantType == null) {
+        if ((variant == null || variantType == null) && !variant.contains("splice")) {
```
onursumer: This will definitely produce a null pointer exception when variant is null. Do we want to return `null` if `variant != null && variant.contains("splice")`? Or do we want to return `null` if `variant == null || variantType == null || variant.contains("splice")`? Or is it something else? :)

inodb: @onursumer doh - good catch fixed

PR #211, opened 2018-08-01
Return p. in hgvspShort output

inodb:
This matches with VEP's HGVSp_Short output so makes it easier to confirm they
are the same.

### DIFF
--- a/web/pom.xml
+++ b/web/pom.xml
@@ -157,8 +157,20 @@
         <filtering>true</filtering>
         <excludes>
           <exclude>*.EXAMPLE</exclude>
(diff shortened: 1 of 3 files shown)


web/pom.xml
```diff
@@ -157,8 +157,20 @@
         <filtering>true</filtering>
         <excludes>
           <exclude>*.EXAMPLE</exclude>
+          <!-- don't alter self signed ssl keys (used for dev) -->
```
inodb: This is useful for testing with self signed certificates so we can test against the live version

PR #220, opened 2018-08-30
My variant info query by dbsnp

inodb:
- query by dbsnpid for myvariant info. Instead of only querying by VariantAnnotation, allow querying directly with given string as well
- switch default myvariant.info to use https (http is currently not working)


web/src/main/java/org/cbioportal/genome_nexus/web/MyVariantInfoController.java
```diff
@@ -32,7 +32,7 @@ public void MyVaraintInfoController(MyVariantInfoService myVariantInfoService)
 
     @ApiOperation(value = "Retrieves myvariant information for the provided list of variants",
         nickname = "fetchMyVariantInfoAnnotationGET")
-    @RequestMapping(value = "/my_variant_info/{variant:.+}",
+    @RequestMapping(value = "/my_variant_info/variant/{variant:.+}",
```
inodb: changed the url to be the same as myvariant.info, in case we want to add more of their endpoints

PR #228, opened 2018-10-10
Add ccdsId to ensembl transcript

inodb:
exact same thing as https://github.com/genome-nexus/genome-nexus/pull/226


inodb:
gonna merge this - it seems to work n it's p much just a copy of the refseq id thing

PR #234, opened 2018-12-05
Query VEP by region

inodb:
- Allow querying of vep by region to use genomenexus VEP wrapper (this does not support querying by hgvs)
- Add gn_vep.region.url property

**NOTES**

~~Querying by region is not part of automated tests, might want to add some mock integration tests using response from the GN VEP wrapper~~ (created new issue for this: https://github.com/genome-nexus/genome-nexus/issues/241)


service/src/main/java/org/cbioportal/genome_nexus/service/annotation/NotationConverter.java
```diff
@@ -154,6 +160,96 @@ else if (ref.length() > 1 || var.length() > 1) {
         return hgvs;
     }
 
+    @Nullable
+    public String genomicToEnsemblRestRegion(GenomicLocation genomicLocation)
```
onursumer: It would be nice to have some tests for non trivial methods like this one.

sandertan:
Hi @inodb , nice PR. Small comment: could you add `gn_vep.region.url` to `application.properties.EXAMPLE`?

Also, do you intend to deprecate the old functionality which uses the Ensembl REST API and `vep.url`, or keep this as an option for people not running their own genome-nexus-vep? Should this be configurable?

service/src/main/java/org/cbioportal/genome_nexus/service/annotation/NotationConverter.java
inodb: added tests

inodb:
@sandertan Thanks for reviewing! I added some info on how to set it in `application.properties.EXAMPLE`. The way it works now, if you don't set `gn_vep.region.url`. It will use `vep.url` as before

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationServiceImpl.java
```diff
@@ -70,6 +74,8 @@
 
     private final CachedVariantAnnotationFetcher cachedExternalResourceFetcher;
     private final CachedVariantIdAnnotationFetcher cachedVariantIdAnnotationFetcher;
+    private final CachedVariantRegionAnnotationFetcher cachedVariantRegionAnnotationFetcher;
```
onursumer: No we have 3 different data fetchers in a single class, and we determine which one to use by looking at the format parameter. This might be another good reason to have separate service implementations for separate formats, since seems like for each format we are using a separate data source. If we have 10 additional formats in the future, we may end up having a long list of fetchers and lengthy switch/case, if/else blocks.

service/src/main/java/org/cbioportal/genome_nexus/service/annotation/VariantAnnotationInputFormat.java
```diff
@@ -0,0 +1,8 @@
+
+package org.cbioportal.genome_nexus.service.annotation;
+
+public enum VariantAnnotationInputFormat {
```
onursumer: We may not even need this `enum`, if we implement a different service for each different data format, and decide which service to call at the controller level.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationServiceImplHgvs.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.commons.logging.LogFactory;
+import org.cbioportal.genome_nexus.service.*;
+import org.cbioportal.genome_nexus.service.annotation.NotationConverter;
+import org.cbioportal.genome_nexus.service.cached.CachedVariantAnnotationFetcher;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Service;
+import org.springframework.beans.factory.annotation.*;
+
+@Service(value="VariantAnnotationServiceImplHgvs")
+public class VariantAnnotationServiceImplHgvs extends BaseVariantAnnotationServiceImpl
```
onursumer: I was wondering if `VariantAnnotationHgvsServiceImpl` or `HgvsVariantAnnotationServiceImpl` is better naming than `VariantAnnotationServiceImplHgvs`. Or, since we already have multiple implementations of the same service and using the `Qualifier`s to pick the right implementation, should we completely remove the `Impl` keyword? (https://octoperf.com/blog/2016/10/27/impl-classes-are-evil/)

web/src/main/java/org/cbioportal/genome_nexus/web/AnnotationController.java
```diff
@@ -58,12 +58,18 @@
...
 public class AnnotationController
 {
-    private final VariantAnnotationService variantAnnotationService;
+    private final VariantAnnotationService hgvsAnnotationService;
+    private final VariantAnnotationService dbsnpAnnotationService;
+    private final GenomicLocationAnnotationService genomicLocationAnnotationService;
 
     @Autowired
-    public AnnotationController(VariantAnnotationService variantAnnotationService)
+    public AnnotationController(@Qualifier("VariantAnnotationServiceImplHgvs") VariantAnnotationService hgvsAnnotationService,
```
onursumer: If we name the implementation class as `HgvsVariantAnnotationService` and the variable as `hgvsVariantAnnotationService`, we may get away without using `@Qualifier("VariantAnnotationServiceImplHgvs")`, and `@Service(value="VariantAnnotationServiceImplHgvs")` at all. Not hundred percent sure if it works though :)

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationServiceImplHgvs.java
inodb: Yeah maybe we can just put it in the same folder n call it e.g. HgvsAnnotationService etc. 

Nice blog reference. We are using a lot of interfaces with a single implementation. Guess we could revisit that

service/src/main/java/org/cbioportal/genome_nexus/service/internal/GenomicLocationAnnotationServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+import org.cbioportal.genome_nexus.service.exception.VariantAnnotationNotFoundException;
+import org.cbioportal.genome_nexus.service.exception.VariantAnnotationWebServiceException;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Service;
+import org.springframework.beans.factory.annotation.*;
+
+import java.util.List;
+
+/**
+ * @author Benjamin Gross
```
onursumer: ...and this one :)

service/src/main/java/org/cbioportal/genome_nexus/service/GenomicLocationAnnotationService.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.cbioportal.genome_nexus.model.GenomicLocation;
+import org.cbioportal.genome_nexus.model.VariantAnnotation;
+import org.cbioportal.genome_nexus.service.exception.VariantAnnotationNotFoundException;
+import org.cbioportal.genome_nexus.service.exception.VariantAnnotationWebServiceException;
+
+import java.util.List;
+
+/**
+ * @author Benjamin Gross
```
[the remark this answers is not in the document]
inodb: good point - thanks! Done

PR #242, opened 2019-01-18
Hotspots by transcript id and protein location

leexgh:
Allow users to Query Genome nexus by transcript id and protein location, return the list of `Hotspots` as the result of the query.

- Fetching by `/cancer_hotspots/transcript/transcriptId`.
Input `transcriptId:string`

- Fetching by  `/cancer_hotspots/proteinLocations`.
Input `[{transcriptId:string,proteinStart:number,proteinEnd:number,mutationType:string}]`
`mutationType` can be one of:  
    `Missense_Mutation`: 3d and single residue
    `In_Frame_Ins`and `In_Frame_Del`: in-frame
    `Splice_Site` and `Splice_Region`: splice

- [x] fetching hotspot by transcript id 
- [x] fetching hotspot by protein location

Future work: 
Implement protein location GET if needed.

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/ProteinLocation.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/ProteinLocation.java
@@ -1,0 +1,82 @@
+package org.cbioportal.genome_nexus.model;
+
+import java.util.*;
+
+public class ProteinLocation
+{
+    private String transcriptId;
+    private Integer start;
+    private Integer end;
+    private String mutationType;
+
+    public String getTranscriptId() {
+        return transcriptId;
+    }
+
+    public Integer getStart() {
+        return start;
+    }
+
+    public Integer getEnd() {
+        return end;
+    }
+
+    public String getMutationType() {
+        return mutationType;
+    }
+
+    public List<Hotspot> filterHotspot(List<Hotspot> hotspots, ProteinLocation proteinLocation){
+        int start = proteinLocation.getStart();
+        int end = proteinLocation.getEnd();
+        String type = proteinLocation.getMutationType();
+        List<Hotspot> result = new ArrayList<>();
+        for (Hotspot hotspot: hotspots) {
+            boolean valid = true;
+            
+            //Position
+            int hotspotStart, hotspotStop;
+            String hotspotPosition = hotspot.getResidue();
+            hotspotPosition = hotspotPosition.replaceAll("[^0-9.]", "");
+            if (hotspotPosition.split("-").length > 1){
+                hotspotStart = Integer.parseInt(hotspotPosition.split("-")[0]);
+                hotspotStop = Integer.parseInt(hotspotPosition.split("-")[1]);
+            }
+            else {
+                hotspotStart = Integer.parseInt(hotspotPosition);
+                hotspotStop = Integer.parseInt(hotspotPosition);
+            }
+            valid &= (start <= hotspotStart && end >= hotspotStop) ? true : false;
+            
+            //Type
+            if (type.equals("Missense_Mutation")) {
+                if (hotspot.getType().contains("3d") || hotspot.getType().contains("single residue")) {
+                    valid &= true;
+                }
+                else {
+                    valid = false;
+                }
+            }
+            if (type.equals("In_Frame_Ins") || type.equals("In_Frame_Del")) {
+                if (hotspot.getType().contains("in-frame")) {
+                    valid &= true;
+                }
+                else {
+                    valid = false;
+                }
+            }
+            if (type.equals("Splice_Site") || type.equals("Splice_Region")) {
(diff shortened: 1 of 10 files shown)


model/src/main/java/org/cbioportal/genome_nexus/model/ProteinLocation.java
```diff
@@ -0,0 +1,91 @@
...
+
+    public Integer getEnd() {
+        return end;
+    }
+
+    public String getMutationType() {
+        return mutationType;
+    }
+
+    public List<Hotspot> filterHotspot(List<Hotspot> hotspots, ProteinLocation proteinLocation){
```
onursumer: This method, and any business logic in general, should be implemented in a service class. In this case, most likely in the `HotspotFilter` class. We try to keep models simple.

model/src/main/java/org/cbioportal/genome_nexus/model/ProteinLocation.java
```diff
@@ -0,0 +1,91 @@
...
+                if (hotspot.getType().contains("splice")) {
+                    valid &= true;
+                }
+                else {
+                    valid = false;
+                }
+            }
+
+            // Add hotspot
+            if (valid) result.add(hotspot);
```
onursumer: Please, never do single line if conditionals. Do this instead:

```suggestion
            if (valid) {
                result.add(hotspot);
            }
```

model/src/main/java/org/cbioportal/genome_nexus/model/ProteinLocation.java
```diff
@@ -0,0 +1,91 @@
...
+        String type = proteinLocation.getMutationType();
+        List<Hotspot> result = new ArrayList<>();
+        for (Hotspot hotspot: hotspots) {
+            boolean valid = true;
+            
+            // Protein location
+            int hotspotStart, hotspotStop;
+            String hotspotPosition = hotspot.getResidue();
+            hotspotPosition = hotspotPosition.replaceAll("[^0-9.]", "");
+            if (hotspotPosition.split("-").length > 1){
```
onursumer: leave a space between `)` and `{`

model/src/main/java/org/cbioportal/genome_nexus/model/ProteinLocation.java
```diff
@@ -0,0 +1,91 @@
...
+                hotspotStart = Integer.parseInt(hotspotPosition.split("-")[0]);
+                hotspotStop = Integer.parseInt(hotspotPosition.split("-")[1]);
+            }
+            else {
+                hotspotStart = Integer.parseInt(hotspotPosition);
+                hotspotStop = Integer.parseInt(hotspotPosition);
+            }
+            valid &= (start <= hotspotStart && end >= hotspotStop) ? true : false;
```
onursumer: ternary operation is redundant here you can simply write this one as:
```suggestion
            valid &= (start <= hotspotStart && end >= hotspotStop);
```

model/src/main/java/org/cbioportal/genome_nexus/model/ProteinLocation.java
```diff
@@ -0,0 +1,91 @@
...
+
+    public List<Hotspot> filterHotspot(List<Hotspot> hotspots, ProteinLocation proteinLocation){
+        int start = proteinLocation.getStart();
+        int end = proteinLocation.getEnd();
+        String type = proteinLocation.getMutationType();
+        List<Hotspot> result = new ArrayList<>();
+        for (Hotspot hotspot: hotspots) {
+            boolean valid = true;
+            
+            // Protein location
```
onursumer: We should have a separate method to parse `hotspotStart` and `hotspotEnd` from hotspot position. Or maybe you can use `Numerical.extractPositiveIntegers` method here?

model/src/main/java/org/cbioportal/genome_nexus/model/ProteinLocation.java
```diff
@@ -0,0 +1,91 @@
...
+                hotspotStop = Integer.parseInt(hotspotPosition.split("-")[1]);
+            }
+            else {
+                hotspotStart = Integer.parseInt(hotspotPosition);
+                hotspotStop = Integer.parseInt(hotspotPosition);
+            }
+            valid &= (start <= hotspotStart && end >= hotspotStop) ? true : false;
+            
+            // Mutation type
+            if (type.equals("Missense_Mutation")) {
```
onursumer: If you write it this way you can avoid all the nested `if/else` blocks:

```suggestion
            validMissense = type.equals("Missense_Mutation") && (hotspot.getType().contains("3d") || hotspot.getType().contains("single residue"));              
```
Then you need to do something like this in the end: `validPosition && (validMissense || validInFrame || ...)`.

leexgh:
Update the comment, implement transcript Id POST.

PR #243, opened 2019-01-23
Make reference genome configurable (GRCh37 or GRCh38)

sandertan:
- This branch refers to data in RC branch of genome-nexus-importer, containing both GRCh37 and GRCh38 data, see https://github.com/genome-nexus/genome-nexus-importer/pull/16
- Environment variables to specify GRCh38 or GRCh37, Ensembl Release and VEP Cache.
- Move Maven commands to Docker
- Include Genome Nexus VEP in docker-compose, see https://github.com/genome-nexus/genome-nexus-vep/pull/1
- Small modifications to java code for compatibility with GRCh38
- Update examples in API documentation

When the changes in genome-nexus-importer and genome-nexus-vep are merged to their respective master branches, the URLs in this PR should be updated.

PR #251, opened 2019-02-22
Add build stage to Dockerfile

ghpxi:


inodb:
Thanks @ghpxi 	! This has just been addressed in: https://github.com/genome-nexus/genome-nexus/pull/243

PR #252, opened 2019-02-25
Fix HotspotFilter issues

zhx828:
- overlap validation should be subset
- valid in-frame test should cover inframe deletion
- Add few more test cases for protiein hotspot annotation

Signed-off-by: Hongxin Zhang <hongxin@cbio.mskcc.org>

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/internal/HotspotFilter.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/internal/HotspotFilter.java
@@ -95,15 +95,18 @@ public class HotspotFilter
             // Protein location
             int hotspotStart = proteinPositionResolver.extractProteinPos(hotspot.getResidue()).getStart();
             int hotspotStop = proteinPositionResolver.extractProteinPos(hotspot.getResidue()).getEnd();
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/internal/HotspotFilter.java
```diff
@@ -95,15 +95,18 @@ else if (hotspotType.contains("3d"))  {
             // Protein location
             int hotspotStart = proteinPositionResolver.extractProteinPos(hotspot.getResidue()).getStart();
             int hotspotStop = proteinPositionResolver.extractProteinPos(hotspot.getResidue()).getEnd();
-            validPosition &= (start <= hotspotStart && end >= hotspotStop);
+
+            // as long as there is overlap, it should be valid position
+            validPosition &= (start <= hotspotStop && end >= hotspotStart);
```
onursumer: I know that we previously approved this but actually `&=` is confusing here, we should just remove the line where we initialize this boolean (` boolean validPosition = true;`) and change this line to:
```suggestion
            boolean validPosition = start <= hotspotStop && end >= hotspotStart;
```

PR #255, opened 2019-02-26
Added PTM endpoints

onursumer:


web/src/main/java/org/cbioportal/genome_nexus/web/param/PtmFilter.java
```diff
@@ -0,0 +1,20 @@
+package org.cbioportal.genome_nexus.web.param;
+
+import io.swagger.annotations.ApiModelProperty;
+
+import java.util.List;
+
+// TODO add other filter params (uniprot, position, type, pubmed id, etc.) when needed
```
onursumer: Not sure if we want to spend time adding more query options at this point. Currently all we need is to query by transcript id.

model/src/main/java/org/cbioportal/genome_nexus/model/PostTranslationalModification.java
```diff
@@ -0,0 +1,87 @@
+package org.cbioportal.genome_nexus.model;
+
+import org.springframework.data.mongodb.core.mapping.Document;
+import org.springframework.data.mongodb.core.mapping.Field;
+
+import java.util.List;
+
+@Document(collection="ptm.experimental")
```
onursumer: data source at dbPTM is named as `experimental and putative PTM sites`, but still unsure if this is the best name for our db collection

component/src/test/java/org/cbioportal/genome_nexus/util/PatternsTest.java
```diff
@@ -0,0 +1,40 @@
...
+    {
+        String[] array = {"cBioPortal", "GenomeNexus", "Make cBioPortal great again", "Also don't forget Genome Nexus"};
+
+        List<Pattern> patterns = Patterns.toStartsWithPatternList(Arrays.asList(array));
+
+        assertTrue(patterns.get(0).matcher("cBioPortal is amazing").matches());
+        assertFalse(patterns.get(0).matcher("cBio").matches());
+        assertFalse(patterns.get(0).matcher("Portal is amazing").matches());
+
+        assertTrue(patterns.get(1).matcher("GenomeNexus is even more amazing").matches());
```
inodb: lol

PR #260, opened 2019-02-27
Variant Format Converters

emz1003:


service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariant.java
```diff
@@ -0,0 +1,43 @@
+package org.cbioportal.genome_nexus.util;
+public class GenomicVariant {
+
+    private static String chromosome;
+    private static Integer start;
+    private static Integer end;
+    private static String ref;
+    private static String alt;
```
inodb: @emz1003 why make these static?

service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariant.java
```diff
@@ -0,0 +1,30 @@
+package org.cbioportal.genome_nexus.util;
+public class GenomicVariant {
+    public GenomicVariantUtil util = new GenomicVariantUtil();
```
inodb: would be good to read up on the concept of util classes: https://stackoverflow.com/questions/25223553/how-can-i-create-an-utility-class. Also there's some people that advocate against using util classes with static methods: https://www.vojtechruzicka.com/avoid-utility-classes/. They propose to just instantiate it once, which could be a better solution

service/src/test/java/org/cbioportal/genome_nexus/util/GenomicVariantTest.java
```diff
@@ -0,0 +1,57 @@
...
+        assertEquals(util.getEnd(), (Integer) 32867862);
+        assertEquals(util.getRef(), "ins");
+        assertEquals(util.getAlt(), "T");
+
+    }
+
+    @Test
+
+    public void testHgvsDeletionToGenomicVariant() {
+        GenomicVariant variant = GenomicVariant.fromHgvs("1:g.19_21del");
```
inodb: this example is actually quite tricky, because you don't know what reference allele is being deleted. Usually you'll see examples like:

```
1:g.19_21delACG
```
and then the reference allele becomes:
```
ACG
```
and alt becomes
```
-
```
or empty string if you prefer

inodb: maybe we can do for the cases where we don't know what the three reference bases deleted were:
```
XXX
```

service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -0,0 +1,42 @@
+package org.cbioportal.genome_nexus.util;
+
+public class GenomicVariantUtil {
+    public GenomicVariant variant = new GenomicVariant();
```
inodb: do u need this field?

service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -0,0 +1,42 @@
...
+        String ans = "";
+        for (String key : keys) {
+            if (hgvs.contains(key))
+                ans = key;
+        }
+        return ans;
+    }
+
+    private static String getRef(String hgvs) {
+        return "XXX";
```
inodb: if this is being used as a constant for missing values, prolly good to add it as a static variable on the class

service/.classpath
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
inodb: i think this file can be removed, maybe add it to `.gitignore`? Something like *.classpath

service/src/main/java/org/cbioportal/genome_nexus/util/minimal_example.in.maf
```diff
@@ -0,0 +1,3 @@
+Chromosome	Start_Position	End_Position	Reference_Allele	Tumor_Seq_Allele1
```
inodb: better maybe to only have this in the `/test/` folder ?

service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -0,0 +1,179 @@
...
+        if (type.equals(Type.SUBSTITUTION))
+            return getPattern("(?<=\\d+)[ATCG]+(?=>)", hgvs);
+        String ans = "";
+        while (ans.length() < getPattern("(?<=[a-z+>])[ATCG]+$", hgvs).length())
+            ans += "X";
+        return ans;
+    }
+
+    private static String getAltFromHgvs(String hgvs, Type type) {
+        if (!type.equals(Type.DELETION))
```
onursumer:
```suggestion
        if (!type.equals(Type.DELETION)) {
            ...
        }
```

service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -0,0 +1,179 @@
...
+            return getPattern("(?<=[a-z+>])[ATCG]+$", hgvs);
+        String alt = getPattern("(?<=[a-z+>])[ATCG]*$", hgvs);
+        if (alt.equals(null))
```
onursumer:
```suggestion
        if (alt.equals(null)) {
            ...
        }
```

service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -0,0 +1,179 @@
...
+            return "";
+        String ans = "";
+        while (ans.length() < alt.length())
```
onursumer:
```suggestion
        while (ans.length() < alt.length()) {
            ...
        }
```

service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -0,0 +1,190 @@
...
+    public static boolean isHgvs(String variant) {
+        return Pattern.matches("^[1-9][0-3]?:[cgmrp]\\.\\d+[ATCG_]\\d*[a-z>]*[ATCG]*$", variant);
+    }
+
+    public static boolean isRegion(String variant) {
+        return Pattern.matches("^\\d{1,2}:\\d+-\\d+:-?1/([ATCG]+|-)$", variant);
+    }
+
+    public static boolean isMafFile(String file) {
+        return Pattern.matches("^[\\w\\-\\./]+.maf$", file);
```
inodb: the mafs often have the extension `txt` or `tsv` as well

service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+        String ans = "";
+        while (ans.length() < getPattern("(?<=[a-z+>])[ATCG]+$", hgvs).length()) {
+            ans += "X";
+        }
+        return ans;
+    }
+
+    private static String getAltFromHgvs(String hgvs, Type type) {
+    if (!type.equals(Type.DELETION)) {
```
onursumer: indentation is not right here

service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -0,0 +1,190 @@
...
+        RefType ref_type = null;
+        Integer start = Integer.parseInt(getPattern("(?<=:)\\d+(?=-)", region));
+        Integer end = Integer.parseInt(getPattern("(?<=-)\\d+(?=:)", region));
+        Type type = null;
+        String ref = null;
+        String alt = getPattern("(?<=:-?1/)[ATCG]+|-$", region);
+        return new GenomicVariant(chr, ref_type, start, end, type, ref, alt);
+    }
+
+    public static String toRegion(GenomicVariant variant){
```
leexgh:
```suggestion
    public static String toRegion(GenomicVariant variant) {
```

service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -0,0 +1,190 @@
...
+        return variant.getChromosome() + ":" + variant.getStart() + "-" + variant.getEnd() + ":1/" + variant.getAlt();
+    }
+
+    public static ArrayList<String> getMafs(String maf_file) {
+        if (!isMafFile(maf_file)){
```
leexgh:
```suggestion
        if (!isMafFile(maf_file)) {
```

service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+        return ans;
+    }
+
+    private static String getAltFromHgvs(String hgvs, Type type) {
+    if (!type.equals(Type.DELETION)) {
+        return getPattern("(?<=[a-z+>])[ATCG]+$", hgvs);
+    }
+        String alt = getPattern("(?<=[a-z+>])[ATCG]*$", hgvs);
+        if (alt.equals(null)){
```
leexgh:
```suggestion
        if (alt.equals(null)) {
```

service/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -0,0 +1,190 @@
...
+            return "";
+        }
+        String ans = "";
+        while (ans.length() < alt.length()) {
+            ans += "X";
+        }
+        return ans;
+    }
+
+    private static RefType getRefTypeFromHgvs (String hgvs) {
```
leexgh:
```suggestion
    private static RefType getRefTypeFromHgvs(String hgvs) {
```

service/src/test/java/org/cbioportal/genome_nexus/util/GenomicVariantTest.java
```diff
@@ -0,0 +1,159 @@
...
+        assertEquals(RefType.GENOMIC, variant.getRefType());
+        assertEquals((Integer) 32867861, variant.getStart());
+        assertEquals((Integer) 32867862, variant.getEnd());
+        assertEquals("X", variant.getRef());
+        assertEquals(Type.INSERTION, variant.getType());
+        assertEquals("T", variant.getAlt());
+    }
+
+    @Test
+
```
leexgh:
```suggestion
```

PR #263, opened 2019-03-07
Update docker-compose instructions: use http://vep:8080 for VEP cache.

### LINKED ISSUE #247: Make sure all end points work with local VEP
When testing the local VEP for variant annotation, only POST /annotation en GET /annotation/{variant} worked. As discussed over Slack, @inodb will take a look at the other 4 endpoints. 

image

gijskant:
Fixes #247.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,11 @@ For example `grch37_ensembl92`, `grch38_ensembl92` or `grch38_ensembl95`:
 export REF_ENSEMBL_VERSION=grch38_ensembl92
 ```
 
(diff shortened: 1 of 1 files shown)


sandertan:
Nice!

inodb:
@gijskant @sandertan nice - so wait there wasn't really a problem, it was just not specifying the right port? Are u sure that the mongo db was empty when testing this?

inodb:
You can remove all data with:
```
docker volume ls | awk '$1 == "local" { print $2 }' | xargs docker volume rm
```

sandertan:
@inodb This fixes my issue the the AnnotationController, which is not using the MongoDB, right?

inodb:
@sandertan it is for caching of the variant data afaik

sandertan:
@inodb Good point. Now I think about it, the issue was that 2 of the 6 annotation endpoints already worked using the previous port settings, so this can't be the solution :(

inodb:
@sandertan 	i think right now it's not possible to use both local vep and an external vep. It might need to use external vep when u using hgvs notation. Or have some way to convert hgvsg to region format. Im not sure if cosmic works either. Easiest solution is prolly to use an external vep for those cases. But use internal vep for all the genomicLocation queries. cBioPortal only uses genomicLocation to interact with GN, so that would suffice for that use case

PR #264, opened 2019-03-08
fix error when transcriptConsequence is null

leexgh:
fix error when transcriptConsequence is null.

### DIFF
--- a/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/CanonicalTranscriptResolver.java
+++ b/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/CanonicalTranscriptResolver.java
@@ -25,13 +25,15 @@ public class CanonicalTranscriptResolver
     {
         List<TranscriptConsequence> transcripts = new ArrayList<>();
 
-        for (TranscriptConsequence transcript : variantAnnotation.getTranscriptConsequences())
-        {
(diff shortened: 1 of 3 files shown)


onursumer:
@leexgh Seems like there is a conflict. We recently moved most of the `@Component`s to a new component module. You probably need to rebase this PR to the latest master and resolve conflicts.

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/CanonicalTranscriptResolver.java
```diff
@@ -25,13 +25,15 @@ public TranscriptConsequence resolve(VariantAnnotation variantAnnotation)
     {
         List<TranscriptConsequence> transcripts = new ArrayList<>();
 
-        for (TranscriptConsequence transcript : variantAnnotation.getTranscriptConsequences())
-        {
-            if (transcript.getTranscriptId() != null &&
-                transcript.getCanonical() != null &&
-                transcript.getCanonical().equals("1"))
+        if (variantAnnotation.getTranscriptConsequences() != null) {
```
inodb: @leexgh personally find it easier to read when `getTranscriptConsequences()` is assigned to a variable and then used in this null check and subsequently to iterate over instead of calling the same method twice

leexgh:
updated, thanks @onursumer @inodb

PR #265, opened 2019-03-13
Add intergenic_consequences to annotation

leexgh:
Add `intergenic_consequences` to annotation. For example:
```
"intergenic_consequences": [
      {
        "variant_allele": "G",
        "consequence_terms": [
          "intergenic_variant"
        ],
        "impact": "MODIFIER"
      }
    ]
```

http://grch37.rest.ensembl.org/vep/human/hgvs/12:g.43130986C%3EG?content-type=application/json

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/IntergenicConsequences.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/IntergenicConsequences.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) 2016 Memorial Sloan-Kettering Cancer Center.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY, WITHOUT EVEN THE IMPLIED WARRANTY OF MERCHANTABILITY OR FITNESS
+ * FOR A PARTICULAR PURPOSE. The software and documentation provided hereunder
+ * is on an "as is" basis, and Memorial Sloan-Kettering Cancer Center has no
+ * obligations to provide maintenance, support, updates, enhancements or
+ * modifications. In no event shall Memorial Sloan-Kettering Cancer Center be
+ * liable to any party for direct, indirect, special, incidental or
+ * consequential damages, including lost profits, arising out of the use of this
+ * software and its documentation, even if Memorial Sloan-Kettering Cancer
+ * Center has been advised of the possibility of such damage.
+ */
+
+/*
+ * This file is part of cBioPortal Genome Nexus.
+ *
+ * cBioPortal is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package org.cbioportal.genome_nexus.model;
+
+import java.util.List;
+
+import org.springframework.data.mongodb.core.mapping.Field;
+
+public class IntergenicConsequences
+{
+    private String impact;
+
+    @Field(value="impact")
+    public String getImpact()
+    {
+        return impact;
+    }
+
+    public void setImpact(String impact)
+    {
+        this.impact = impact;
+    }
+
+    private String variant_allele;
+
+    @Field(value="variant_allele")
+    public String getVariantAllele()
+    {
+        return variant_allele;
+    }
+
+    public void setVariantAllele(String variant_allele)
+    {
+        this.variant_allele = variant_allele;
+    }
+
+    private List<String> consequence_terms;
+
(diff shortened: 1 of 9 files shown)


model/src/main/java/org/cbioportal/genome_nexus/model/IntergenicConsequences.java
```diff
@@ -0,0 +1,80 @@
...
+    {
+        return impact;
+    }
+
+    public void setImpact(String impact)
+    {
+        this.impact = impact;
+    }
+
+    private String variant_allele;
```
onursumer: No java variables with `_` please :)

```suggestion
    private String variantAllele;
```

model/src/main/java/org/cbioportal/genome_nexus/model/IntergenicConsequences.java
```diff
@@ -0,0 +1,80 @@
...
+    {
+        return variant_allele;
+    }
+
+    public void setVariantAllele(String variant_allele)
+    {
+        this.variant_allele = variant_allele;
+    }
+
+    private List<String> consequence_terms;
```
onursumer: same as above
```suggestion
    private List<String> consequenceTerms;
```

service/src/main/java/org/cbioportal/genome_nexus/service/mixin/IntergenicConsequencesMixin.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class IntergenicConsequencesMixin
+{
+    @JsonProperty(value="impact", required = true)
+    private String impact;
+
+    @JsonProperty(value="variant_allele", required = true)
+    private String variant_allele;
```
onursumer:
```suggestion
    private String variantAllele;
```

service/src/main/java/org/cbioportal/genome_nexus/service/mixin/IntergenicConsequencesMixin.java
```diff
@@ -0,0 +1,19 @@
...
+
+    @JsonProperty(value="consequence_terms", required = true)
+    private List<String> consequence_terms;
```
onursumer:
```suggestion
    private List<String> consequenceTerms;
```

service/src/main/java/org/cbioportal/genome_nexus/service/mixin/IntergenicConsequencesMixin.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class IntergenicConsequencesMixin
+{
+    @JsonProperty(value="impact", required = true)
+    private String impact;
+
+    @JsonProperty(value="variantAllele", required = true)
```
inodb: good to keep the original vep field for the user. that's what we do with the other vep mixin classes as well (i.e. in this case `variant_allele`)

inodb: check if this works with an empty database. This mixin should map to the original JSON response from VEP

inodb: We just tested this together :) - it only works after being cached once (because the whole response is stored inside the database and the database model does map correctly)

Filed an issue, we should prolly get rid of service mixins completely: https://github.com/genome-nexus/genome-nexus/issues/268

model/src/main/java/org/cbioportal/genome_nexus/model/VariantAnnotation.java
```diff
@@ -61,6 +61,7 @@
     private Integer strand;
     private String mostSevereConsequence;
     private List<ColocatedVariant> colocatedVariants;
+    private List<IntergenicConsequences> intergenic_consequences;
```
inodb: @leexgh i guess this works now to make the integration test pass, but it is not consistent with the other member names. Would be better to change the integration test instead of changing the model to fit the test

web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
```diff
@@ -0,0 +1,67 @@
...
+        String[] variant = {
+            "12:g.43130875C>G",
+            "12:g.43130996C>G"
+        };
+
+        //////////////////
+        // GET requests //
+        //////////////////
+
+        VariantAnnotation ann0 = this.fetchVariantAnnotationGET(variant[0]);
```
leexgh: Most fields are null in `ann0` because of different mapping names. It can only return the fields that have the same mapping name, such as `start` and `end`. We may fix issue #268 and add more tests here.

model/src/main/java/org/cbioportal/genome_nexus/model/VariantAnnotation.java
leexgh: @inodb yes it is, I already change the test and model name to keep consistent. Thanks!

web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
onursumer: ...or we configure the integration tests to use mixins to get the correct mapping :)

PR #267, opened 2019-04-13
Add gnomad data into myVariantInfo endpoint

leexgh:
Update myVairiantInfo endpoint, adding gnomad data (allele count, allele number, number of homozygotes, allele frequency) into the myVariantInfo endpoint.

https://myvariant.info/v1/variant/chr10%3Ag.104264107C%3ET

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/my_variant_info_model/AlleleCount.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/my_variant_info_model/AlleleCount.java
@@ -5,130 +5,121 @@ import org.springframework.data.mongodb.core.mapping.Field;
 public class AlleleCount
 {
 
-    // allele count in total
     @Field(value = "ac")
-    private Integer ac;
+    private Integer alleleCount;
 
-    // allele count African
     @Field(value = "ac_afr")
-    private Integer acAfr;
+    private Integer alleleCountAfrican;
 
-    // allele count Latino
     @Field(value = "ac_amr")
-    private Integer acAmr;
+    private Integer alleleCountLatino;
 
-    // allele count Ashkenazi Jewish
     @Field(value = "ac_asj")
-    private Integer acAsj;
+    private Integer alleleCountAshkenaziJewish;
 
-    // allele count East Asian
     @Field(value = "ac_eas")
-    private Integer acEas;
+    private Integer alleleCountEastAsian;
 
-    // allele count European(Finnish)
     @Field(value = "ac_fin")
-    private Integer acFin;
+    private Integer alleleCountEuropeanFinnish;
 
-    // allele count European(non-Finnish)
     @Field(value = "ac_nfe")
-    private Integer acNfe;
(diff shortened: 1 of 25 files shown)


model/src/main/java/org/cbioportal/genome_nexus/model/my_variant_info_model/AlleleFrequency.java
```diff
@@ -0,0 +1,134 @@
+package org.cbioportal.genome_nexus.model.my_variant_info_model;
+
+import org.springframework.data.mongodb.core.mapping.Field;
+
+public class AlleleFrequency
+{
+
+    // allele frequency in total
+    @Field(value = "af")
+    private Float af;
```
onursumer: I think we should use `Double` instead of `Float` to be consistent with the other models.

model/src/main/java/org/cbioportal/genome_nexus/model/my_variant_info_model/AlleleCount.java
```diff
@@ -0,0 +1,134 @@
+package org.cbioportal.genome_nexus.model.my_variant_info_model;
+
+import org.springframework.data.mongodb.core.mapping.Field;
+
+public class AlleleCount
+{
+
+    // allele count in total
+    @Field(value = "ac")
+    private Integer ac;
```
onursumer: Ideally we should name the variables like `alleleCount` instead of `ac`, `alleleCountAfrican` instead of `acAfr` etc. In that case wouldn't even need to add any comments since the variable names would be self descriptive.

model/src/main/java/org/cbioportal/genome_nexus/model/my_variant_info_model/Gnomad.java
```diff
@@ -0,0 +1,64 @@
+package org.cbioportal.genome_nexus.model.my_variant_info_model;
+
+import org.springframework.data.mongodb.core.mapping.Field;
+
+public class Gnomad
+{
+
+    // Allele count
```
onursumer: the variable names are perfect in this class, so the comments are redundant.

leexgh:
@onursumer update, thanks!

web/src/test/java/org/cbioportal/genome_nexus/web/MyVariantInfoIntegrationTest.java
```diff
@@ -48,25 +61,44 @@ public void testMyVariantInfoAnnotation()
         // GET requests //
         //////////////////
 
-        MyVariantInfo ann0 = this.fetchMyVariantInfoAnnotationGET(variants[0]);
-        // mutdb annotation should be 7
-        assertEquals("7", ann0.getMutdb().getChrom());
+        String chrom = this.fetchMyVariantInfoAnnotationGET(variants[0]).get("mutdb").get("chrom").toString();
```
inodb: i like it 👍 - very readable

PR #271, opened 2019-05-13
Fixed incorrect filter conditions

onursumer:


inodb:
@leexgh me and hongxin reviewed it - oops. This happens. Onur spotted it in IntelliJ. There it gave some notification that there was an error in that logical condition. There is probably some way to add these warnings to the continuous integration testing to avoid these type of errors in the future. Could be worth investigating

PR #274, opened 2019-05-20
Fix myVariantInfo model

leexgh:
### Different structures
The gene field from myVariantInfo has two types:
1. list
https://myvariant.info/v1/variant/chr17%3Ag.41276045_41276046del
```
    "gene": [
      {
        "geneid": "10230",
        "symbol": "NBR2"
      },
      {
        "geneid": "672",
        "symbol": "BRCA1"
      }
    ],
``` 
2. object
https://myvariant.info/v1/variant/chr17%3Ag.7577121G%3EA
```
    "gene": {
      "geneid": "7157",
      "symbol": "TP53"
    },
```
Some variants may get errors if we set to a particular type. So we remove the `gene` for now since we are not using it. 

We may add it back if myVariantInfo is using a unified structure or we need this field in the future.

**UPDATE:**
MyVariantInfo provides a parameter to explicitly enforce `gene` field to be a list/array.
https://myvariant.info/v1/variant/chr17%3Ag.7577121G%3EA?fields=dbsnp&always_list=dbsnp.gene

"always_list" parameter accepts a list of fields (sep by commas) to ensure these fields return as a list.

### Use myVariantInfo variant id structure for `del` and `ins`.

- del example: 17:g.41276045_41276046del
- ins example: 2:g.17142_17143insA
- delins example: 13:g.28611219_28611221delinsGC (this variant is not available now, should be available on early June - myVariantInfo new release)


service/src/main/java/org/cbioportal/genome_nexus/service/internal/MyVariantInfoServiceImpl.java
```diff
@@ -123,6 +123,17 @@ private String buildRequest(String variant)
         {
             sb.insert(0,"chr");
         }
+        // Check if the variant id contains "del", if so we will remove the deleted nucleotides
```
onursumer: Can we separate this to a utility class, and maybe add some tests for it?

onursumer: Should we separate this logic into a utility method (under component module) and add some unit tests too?

leexgh: @onursumer done

component/src/main/java/org/cbioportal/genome_nexus/util/VariantId.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class VariantId
+{
+    /**
+     * Build variant id in myVariantInfo id structure
+     *
+     * @param input input variant
+     * @return  variant id
+     */
+    public static String buildVariantId(String input)
```
inodb: @leexgh The reason for the utility function is to make it general i.e. not specific to `MyVariantInfo`. There are two things that happen in this function:

1. Adding `chr`. Some genome builds indicate chromosomes without `chr` others without
2. Normalize `delACGinsT` to `delinsT` as per the Hgvs specification

then `buildRequest` could call the utility functions of HgvsUtil (or some name like that, since these are utilities related to Hgvs formatting):

```
String addChrPrefix(String hgvs)
String removeDeletedBases(String hgvs)
```

That way you can re-use those last two functions

leexgh:
@inodb done, thanks!

inodb:
@leexgh Nice work 🎉 	!

PR #280, opened 2019-06-13
fix annotation_summary slowness

leexgh:
For example:
400 mutations, reduce consuming time from 3m15s to 0.4s (local)

### DIFF
--- a/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/EnsemblRepositoryCustom.java
+++ b/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/EnsemblRepositoryCustom.java
@@ -31,6 +31,8 @@
 
 package org.cbioportal.genome_nexus.persistence.internal;
 
+import java.util.HashMap;
+
 import org.cbioportal.genome_nexus.model.EnsemblGene;
 import org.cbioportal.genome_nexus.model.EnsemblTranscript;
 
@@ -39,4 +41,5 @@ public interface EnsemblRepositoryCustom
     EnsemblTranscript findOneByHugoSymbolIgnoreCase(String hugoSymbol, String isoformOverrideSource);
     EnsemblGene getCanonicalEnsemblGeneIdByHugoSymbol(String hugoSymbol);
     EnsemblGene getCanonicalEnsemblGeneIdByEntrezGeneId(String ensemblGeneId);
+    HashMap<String, String> getHugoSymbolToEntrezGeneIdMap();
 }
--- a/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/EnsemblRepositoryCustom.java
+++ b/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/EnsemblRepositoryCustom.java
@@ -31,8 +31,6 @@
 
 package org.cbioportal.genome_nexus.persistence.internal;
 
-import java.util.HashMap;
-
 import org.cbioportal.genome_nexus.model.EnsemblGene;
 import org.cbioportal.genome_nexus.model.EnsemblTranscript;
 
@@ -41,5 +39,5 @@ public interface EnsemblRepositoryCustom
     EnsemblTranscript findOneByHugoSymbolIgnoreCase(String hugoSymbol, String isoformOverrideSource);
     EnsemblGene getCanonicalEnsemblGeneIdByHugoSymbol(String hugoSymbol);
     EnsemblGene getCanonicalEnsemblGeneIdByEntrezGeneId(String ensemblGeneId);
-    HashMap<String, String> getHugoSymbolToEntrezGeneIdMap();
+    String findEntrezGeneIdByHugoSymbol(String hugoSymbol);
 }
--- a/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/EnsemblRepositoryImpl.java
+++ b/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/EnsemblRepositoryImpl.java
@@ -8,6 +8,8 @@ import org.springframework.data.mongodb.core.MongoTemplate;
 import org.springframework.data.mongodb.core.query.Criteria;
 import org.springframework.data.mongodb.core.query.Query;
 
+import java.util.HashMap;
+import java.util.List;
 import java.util.regex.Pattern;
 
 import com.mongodb.BasicDBList;
@@ -31,6 +33,7 @@ public class EnsemblRepositoryImpl implements EnsemblRepositoryCustom
 
     private EnsemblCanonical findOneCanonicalByHugoSymbol(String hugoSymbol) {
         EnsemblCanonical ensemblCanonical;
+        // mark
         Query query;
 
         // check approved symbols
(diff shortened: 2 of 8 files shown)


persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/EnsemblRepositoryImpl.java
```diff
@@ -109,4 +111,34 @@ public EnsemblTranscript findOneByHugoSymbolIgnoreCase(String hugoSymbol, String
         }
         return null;
     }
+
+    @Override
+    public HashMap<String, String> getHugoSymbolToEntrezGeneIdMap() {
```
onursumer: Every time this method is called the `HashMap` is constructed over and over again. Can we rename this method to `initHugoSymbolToEntrezGeneIdMap`, make it private, and invoke this method only in the constructor? Ideally we should only expose (make public) the method `findEntrezGeneIdByHugoSymbol(String hugoSymbol)`.

persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/EnsemblRepositoryImpl.java
```diff
@@ -109,4 +111,34 @@ public EnsemblTranscript findOneByHugoSymbolIgnoreCase(String hugoSymbol, String
...
+        for (EnsemblCanonical transcript : transcripts) {
+            String[] previousSymbols = transcript.getPreviousSymbols();
+            String[] synonyms = transcript.getSynonyms();
+
+            if (!map.containsKey(transcript.getHugoSymbol())) {
+                map.put(transcript.getHugoSymbol(), transcript.getEntrezGeneId());
+            }
+            if (previousSymbols != null) {
+                for (String previousSymbol : previousSymbols) {
+                    if (!map.containsKey(previousSymbol)) {
```
onursumer: you don't really need to check if it contains the key, if it contains the key already it will override with the new (hopefully the same) value. This check adds additional complexity to the code, and we don't even handle the case when it contains different values for the same symbols.

persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/EnsemblRepositoryImpl.java
```diff
@@ -109,4 +111,34 @@ public EnsemblTranscript findOneByHugoSymbolIgnoreCase(String hugoSymbol, String
...
+                        map.put(previousSymbol, transcript.getEntrezGeneId());
+                    }
+                }
+            }
+            if (synonyms != null) {
+                for (String synonym : synonyms) {
+                    if (!map.containsKey(synonym)) {
```
onursumer: same comment as above. We don't take any action if there is already an entry for the same symbol, so we can simply override it for now.

service/src/main/java/org/cbioportal/genome_nexus/service/EnsemblService.java
```diff
@@ -27,4 +28,5 @@ EnsemblTranscript getCanonicalEnsemblTranscriptByHugoSymbol(String hugoSymbol, S
 
     List<EnsemblTranscript> getEnsemblTranscripts(String geneId, String proteinId, String hugoSymbol);
     List<EnsemblTranscript> getEnsemblTranscripts(List<String> transcriptIds, List<String> geneIds, List<String> proteinIds, List<String> hugoSymbols);
+    HashMap<String, String> getHugoSymbolToEntrezGeneIdMap();
```
onursumer: Similar to the repository interface let's add a method like `getEntrezGeneIdByHugoSymbol(String hugoSymbol)`

service/src/main/java/org/cbioportal/genome_nexus/service/annotation/EntrezGeneIdResolver.java
```diff
@@ -15,12 +17,14 @@
 {
     private final CanonicalTranscriptResolver canonicalTranscriptResolver;
     private final EnsemblService ensemblService;
+    private final HashMap<String, String> hugoSymbolToEntrezGeneIdMap;
```
onursumer: remove this one

service/src/main/java/org/cbioportal/genome_nexus/service/annotation/EntrezGeneIdResolver.java
```diff
@@ -15,12 +17,14 @@
...
 
     public EntrezGeneIdResolver(CanonicalTranscriptResolver canonicalTranscriptResolver,
                                 EnsemblService ensemblService)
     {
         this.canonicalTranscriptResolver = canonicalTranscriptResolver;
         this.ensemblService = ensemblService;
+        this.hugoSymbolToEntrezGeneIdMap = ensemblService.getHugoSymbolToEntrezGeneIdMap();
```
onursumer: remove this one

service/src/main/java/org/cbioportal/genome_nexus/service/annotation/EntrezGeneIdResolver.java
```diff
@@ -32,18 +36,20 @@ public String resolve(TranscriptConsequence transcriptConsequence) throws Ensemb
...
+            //     ensemblGene = this.ensemblService
+            //             .getCanonicalEnsemblGeneIdByHugoSymbol(transcriptConsequence.getGeneSymbol());
+            // } catch (NoEnsemblGeneIdForHugoSymbolException e) {
+            //     // ignore silently
+            // }
+
+            // if (ensemblGene != null) {
+            //     entrezGeneId = ensemblGene.getEntrezGeneId();
+            // } 
+            entrezGeneId = this.hugoSymbolToEntrezGeneIdMap.get(transcriptConsequence.getGeneSymbol());
```
onursumer: replace this with:

```suggestion
            entrezGeneId = this.ensemblService.getEntrezGeneIdByHugoSymbol(transcriptConsequence.getGeneSymbol());
```

service/src/main/java/org/cbioportal/genome_nexus/service/internal/EnsemblServiceImpl.java
```diff
@@ -168,4 +169,9 @@ else if (hugoSymbols != null && hugoSymbols.size() > 0) {
             return new ArrayList<EnsemblTranscript>();
         }
     }
+
+    @Override
+    public HashMap<String, String> getHugoSymbolToEntrezGeneIdMap() {
```
onursumer: implement a function like `getEntrezGeneIdByHugoSymbol(String hugoSymbol)` instead of this one

persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/EnsemblRepositoryCustom.java
```diff
@@ -39,4 +41,5 @@
     EnsemblTranscript findOneByHugoSymbolIgnoreCase(String hugoSymbol, String isoformOverrideSource);
     EnsemblGene getCanonicalEnsemblGeneIdByHugoSymbol(String hugoSymbol);
     EnsemblGene getCanonicalEnsemblGeneIdByEntrezGeneId(String ensemblGeneId);
+    HashMap<String, String> getHugoSymbolToEntrezGeneIdMap();
```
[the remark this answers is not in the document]
inodb: It might make sense to map to the whole `EnsemblCanonical` transcript instead of only `EntrezGeneId`. That way you can get the other info about the canonical transcript in a quick way as well if you would want to. Anyway I know this PR is a quick fix, so prolly better to leave it as is for now

leexgh:
@onursumer fixed, thanks!

PR #284, opened 2019-06-20
Updated the logic for resolveHgvspShortFromAAs

### LINKED ISSUE #282: Annotation summary fails for the variant 5:g.112174750_112174751insGAA
http://genomenexus.org/annotation/5%3Ag.112174750_112174751insGAA?fields=hotspots%2Cmutation_assessor%2Cannotation_summary

`java.lang.StringIndexOutOfBoundsException: String index out of range: 2` at https://github.com/genome-nexus/genome-nexus/blob/052c65652a9e2092e98477c6f6200e4875652c5d/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/ProteinChangeResolver.java#L178

onursumer:
Fix #282


inodb:
@onursumer might be good to also do some `try/catch` that would avoid breaking the whole POST, but could be separate PR

PR #298, opened 2019-07-30
fix annotation fail

leexgh:
The annotation will fail if one of the variants is failed.

we catch all exceptions, ignore if they fail and just continue with the next one.


inodb:
the failure was not related to this PR

PR #320, opened 2019-09-04
Fix myvariant.info when using `gn_vep.region.url`

inodb:
In this case the id for VariantAnnotation is region instead of hgvsg, so
instead we determine hgvsg from the VEP response.

### DIFF
--- a/web/src/main/java/org/cbioportal/genome_nexus/web/mixin/TranscriptConsequenceMixin.java
+++ b/web/src/main/java/org/cbioportal/genome_nexus/web/mixin/TranscriptConsequenceMixin.java
@@ -21,6 +21,10 @@ public class TranscriptConsequenceMixin
     @ApiModelProperty(value = "HGVSc", required = false)
     private String hgvsc;
 
+    @JsonProperty(value="hgvsg", required = true)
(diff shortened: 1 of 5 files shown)


web/src/main/java/org/cbioportal/genome_nexus/web/mixin/TranscriptConsequenceMixin.java
```diff
@@ -21,6 +21,10 @@
     @ApiModelProperty(value = "HGVSc", required = false)
     private String hgvsc;
 
+    @JsonProperty(value="hgvsg", required = true)
```
onursumer: Also add an annotation for this into `service/mixin/TranscriptConsequenceMixin.java` for consistency? I guess this still works because we always read from the cache. In case we have to read directly from the web service this field may be ignored. There is a `@JsonIgnoreProperties(ignoreUnknown = true)` in `service/mixin/TranscriptConsequenceMixin.java`.

inodb: thanks - done

PR #327, opened 2019-10-02
Pull genome-nexus-frontend using jitpack

inodb:
- Update pom to copy static resources
- Remove existing static files

TODO:
- [x] certain font assets and icons are missing when compiling with jitpack (this turned out to be a maven filtering issue)
- [x] only works when compiling twice, should work immediately

TODO separate issues:

- Make API URL configurable (it now always uses https://www.genomenexus.org). When included in the genome-nexus backend war it should use that backend instead (https://github.com/genome-nexus/genome-nexus-frontend/issues/32)
- URL routing, should always go to index page (https://github.com/genome-nexus/genome-nexus/issues/328)


leexgh:
@inodb thank you so much! finally understand where i was stuck!

PR #351, opened 2019-11-07
Fix mutation assessor enrichment failure

### LINKED ISSUE #350: X,1325430,1325430,G,A gives 404
The real error is with Mutation Assessor, but the whole request should not fail because of it. Just the mutation_assessor part should be empty

inodb:
Fix #350 

Mutation assessor fails for X,1325430,1325430,G,A. Ignore that exception
and log as a warning instead.

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/internal/VEPEnrichmentService.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/internal/VEPEnrichmentService.java
@@ -32,32 +32,39 @@
 
 package org.cbioportal.genome_nexus.service.internal;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.cbioportal.genome_nexus.model.VariantAnnotation;
 import org.cbioportal.genome_nexus.service.AnnotationEnricher;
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/internal/VEPEnrichmentService.java
```diff
@@ -32,32 +32,48 @@
...
+                Entry<String, AnnotationEnricher> enricher = it.next();
+                try {
+                    enricher.getValue().enrich(variantAnnotation);
+                } catch (Exception e) {
+
+                    LOG.warn("Failed to enrich with " + enricher.getKey() + ": " + variantAnnotation.getVariant() + " " + e.getLocalizedMessage());
+                }
+
+            }
             for (AnnotationEnricher enricher: enrichers.values())
```
onursumer: @inodb Shouldn't we remove this `for` loop, since now we moved the enrichment to the `while` loop above?

inodb: yes sorry just remove it 👍

Issue #156, opened 2018-03-28
Display and/or specify Ensembl Release and Reference Genome


sandertan:
I was wondering if it will be possible to display and/or specify the Ensembl release and reference genome used.

inodb:
@sandertan where? In cBioPortal or in the output from the API?

sandertan:
@inodb Hi Ino, I think it would be useful to query the API for both hg19 and hg38. I understand this would be a lot of work to implement, but we sometimes hear that users would like hg38 for their cBioPortal data. Also, Kelsey is working on a PR to cBioPortal to support multiple reference genomes for mutations https://github.com/cBioPortal/cbioportal/pull/3703 .

Regarding the output, I can imagine it would be useful to display the used Ensembl Release and the reference genome. This would explain the context of the returned result.

inodb:
Hi Sander,

Yeah we were thinking to run a hg38 instance as well at some point. Similar
to how ensembl has multiple instances. It would prolly just be passing some
parameter to genome-nexus-importer and setting some config for
genome-nexus. Then hg19 or hg38 would be in the URL.

PR #361, opened 2019-12-31
sort variant requests before making chunks

sheridancbio:
- base class BaseCachedExternalResourceFetcher chunk logic now common to all
- sort occurs on full set of requests before chunks are made
- removed the FIXED_ENTRY prepended to each chunk : (see genome-nexus issue #156)
    - tests showed that the original problem is no longer happening, so workaround is not needed and is degrading performance

Co-authored-by: Manda Wilson <1458628+mandawilson@users.noreply.github.com>
Co-authored-by: Robert Sheridan <7747489+sheridancbio@users.noreply.github.com>

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/internal/RegionVariantAnnotationService.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/internal/RegionVariantAnnotationService.java
@@ -44,7 +44,7 @@ import org.springframework.beans.factory.annotation.*;
 @Service
 public class RegionVariantAnnotationService extends BaseVariantAnnotationServiceImpl
 {
-    private static final Log LOG = LogFactory.getLog(DbsnpVariantAnnotationService.class);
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/internal/RegionVariantAnnotationService.java
```diff
@@ -44,7 +44,7 @@
 @Service
 public class RegionVariantAnnotationService extends BaseVariantAnnotationServiceImpl
 {
-    private static final Log LOG = LogFactory.getLog(DbsnpVariantAnnotationService.class);
+    private static final Log LOG = LogFactory.getLog(RegionVariantAnnotationService.class);
```
onursumer: good catch!

inodb:
@sheridancbio @mandawilson is this ok to merge?

PR #365, opened 2020-01-07
Remove old auto VCF -> MAF conversion code

inodb:
This in some cases results in incorrect annotations, so better to have
people explicitly convert their VCF to MAF before importing by e.g.
using:

https://github.com/genome-nexus/genome-nexus-cli#convert-vcf-to-maf-ready-for-annotation

One thing is that the annotation pipeline does somehow correctly annotate it. I'm not entirely sure why, I guess it uses a different endpoint then:

https://www.genomenexus.org/annotation/genomic/7%2C55242467%2C55242482%2CAATTAAGAGAACATCT%2CAGCAA?isoformOverrideSource
https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/113

### DIFF
--- a/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java
+++ b/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java
@@ -94,11 +94,7 @@ public class NotationConverter
     }
 
     /*
-     * Normalize genomic location:
-     *
-     * 1. Convert VCF style start,end,ref,alt to MAF by looking for common
-     * prefix. (TODO: not sure if this is always a good idea)
-     * 2. Normalize chromsome names.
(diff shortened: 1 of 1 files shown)


inodb:
Actually looked into this some more and that conversion is fine in this case (albeit a little confusing):

this and this gives the same response (apart ofc from the id):

http://grch37.rest.ensembl.org/vep/human/hgvs/7:g.55242467_55242486delinsAGCAA?content-type=application/json&hgvs=1'

http://grch37.rest.ensembl.org/vep/human/hgvs/7:g.55242468_55242486delinsGCAA?content-type=application/json&hgvs=1'

There is another problem when trying to relate the response of the post request to the input. Since the variants get sorted you can't actually relate your input to the output when the genomic location gets adjusted

PR #371, opened 2020-02-04
Map All Variants in POST Request to a Response VariantAnnotation Object 

averyniceday:
- expand model for annotation success flag
- change logic to include failed variants for post
- change logic to return response including duplicates
- (e.g POST request w/ same variant twice, will return both in returned list)
- change logic to include failed variant for GET
- remove validation for path variable for GET

### DIFF
--- a/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java
+++ b/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java
@@ -297,8 +297,14 @@ public class NotationConverter
     public List<String> genomicToHgvs(List<GenomicLocation> genomicLocations)
     {
         List<String> hgvsList = new ArrayList<>();
-        hgvsList.addAll(this.genomicToHgvsMap(genomicLocations).keySet());
-
+        
+        for (GenomicLocation location : genomicLocations) { 
+            String hgvs = this.genomicToHgvs(location);
+            if (hgvs != null) {
+                hgvsList.add(hgvs);
+            }
+        }
+        
         return hgvsList;
     }
 
@@ -322,7 +328,12 @@ public class NotationConverter
     public List<String> genomicToEnsemblRestRegion(List<GenomicLocation> genomicLocations)
     {
         List<String> ensemblRestRegionsList = new ArrayList<>();
-        ensemblRestRegionsList.addAll(this.genomicToEnsemblRestRegionMap(genomicLocations).keySet());
+        for (GenomicLocation location : genomicLocations) { 
+            String ensemblRestRegion = this.genomicToEnsemblRestRegion(location);
+            if (ensemblRestRegion != null) {
+                ensemblRestRegionsList.add(ensemblRestRegion);
+            }
+        }
 
         return ensemblRestRegionsList;
     }
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/VariantAnnotation.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/VariantAnnotation.java
@@ -63,6 +63,7 @@ public class VariantAnnotation
     private List<ColocatedVariant> colocatedVariants;
     private List<IntergenicConsequences> intergenicConsequences;
     private List<TranscriptConsequence> transcriptConsequences;
+    private Boolean successfullyAnnotated;
 
     private MutationAssessorAnnotation mutationAssessorAnnotation;
     private MyVariantInfoAnnotation myVariantInfoAnnotation;
@@ -88,6 +89,7 @@ public class VariantAnnotation
         this.variant = variant;
         this.annotationJSON = annotationJSON;
         this.dynamicProps = new LinkedHashMap<>();
+        this.successfullyAnnotated = !(annotationJSON == null);
     }
 
     public String getVariant()
@@ -272,20 +274,18 @@ public class VariantAnnotation
     }
 
     public String getHgvsg() {
-        if (this.getVariantId().contains("g."))
+        if (this.getVariantId() != null && this.getVariantId().contains("g."))
         {
             return this.getVariantId();
-        }
-        // id is not of hgvsg format
-        else {
-            // determine hgvsg from VEP output
+        } else if (this.getTranscriptConsequences() == null) {
+            return null;
+        } else {
+            // id is not of hgvsg format
             for (TranscriptConsequence ts : this.getTranscriptConsequences()) {
                 if (ts.getHgvsg() != null && !ts.getHgvsg().isEmpty()) {
                     return ts.getHgvsg();
                 }
-
             }
-            // TODO: check intergenic if still no hgvsg
             return null;
         }
     }
@@ -300,6 +300,14 @@ public class VariantAnnotation
 
     public void setOncokbAnnotation(OncokbAnnotation oncokbAnnotation) {
         this.oncokbAnnotation = oncokbAnnotation;
+    }
+
+    public void setSuccessfullyAnnotated(Boolean successfullyAnnotated) {
+        this.successfullyAnnotated = successfullyAnnotated;
+    }
+
+    public Boolean isSuccessfullyAnnotated() {
+        return successfullyAnnotated;
     }
 
     public void setDynamicProp(String key, Object value)
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java
@@ -14,7 +14,6 @@ import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.data.mongodb.repository.MongoRepository;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.HttpServerErrorException;
-
 import java.util.*;
 import java.util.stream.Collectors;
 
@@ -81,14 +80,12 @@ public abstract class BaseCachedExternalResourceFetcher<T, R extends MongoReposi
                 "Will not attempt to store variant in Mongo database.");
             saveRawValue = false;
         }
-
         if (!instance.isPresent())
         {
             // get the annotation from the web service and save it to the DB
             try {
                 // get the raw annotation string from the web service
                 DBObject rawValue = this.fetcher.fetchRawValue(id);
-
                 // construct an instance to return:
                 // this does not contain all the information obtained from the web service
                 // only the fields mapped to the VariantAnnotation model will be returned
@@ -121,7 +118,7 @@ public abstract class BaseCachedExternalResourceFetcher<T, R extends MongoReposi
         }
     }
 
-    public List<T> fetchAndCache(List<String> ids) throws ResourceMappingException
+    public Map<String, T> constructFetchedMap(List<String> ids) throws ResourceMappingException
     {
         boolean saveValues = true;
         Set<String> uniqueIds = new LinkedHashSet<>(ids);
@@ -150,13 +147,17 @@ public abstract class BaseCachedExternalResourceFetcher<T, R extends MongoReposi
 
         // also remove invalid ids
         needToFetch = needToFetch.stream().filter(this::isValidId).collect(Collectors.toSet());
-
         // fetch missing instances
         if (needToFetch.size() > 0) {
             // get the annotation from the web service and save it to the DB
             this.fetchAndCache(needToFetch, idToInstance, saveValues);
         }
+        return idToInstance;
+    }
 
+    public List<T> fetchAndCache(List<String> ids) throws ResourceMappingException
+    {
+        Map<String, T> idToInstance = constructFetchedMap(ids);
         Collection<T> values = idToInstance.values();
         values.removeIf(Objects::isNull);
 
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java
@@ -1,0 +1,102 @@
+package org.cbioportal.genome_nexus.service.cached;
+
+import com.mongodb.DBObject;
+import org.cbioportal.genome_nexus.model.VariantAnnotation;
+import org.cbioportal.genome_nexus.persistence.VariantAnnotationRepository;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.cbioportal.genome_nexus.persistence.GenericMongoRepository;
+import org.cbioportal.genome_nexus.service.CachedExternalResourceFetcher;
+import org.cbioportal.genome_nexus.service.ExternalResourceFetcher;
+import org.cbioportal.genome_nexus.service.ResourceTransformer;
+import org.cbioportal.genome_nexus.service.exception.ResourceMappingException;
+import org.cbioportal.genome_nexus.util.NaturalOrderComparator;
+import org.springframework.dao.DataAccessResourceFailureException;
+import org.springframework.dao.DataIntegrityViolationException;
+import org.springframework.data.mongodb.repository.MongoRepository;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.HttpServerErrorException;
+import java.util.*;
+import java.util.stream.Collectors;
+
+public abstract class BaseCachedVariantAnnotationFetcher
+    extends BaseCachedExternalResourceFetcher<VariantAnnotation, VariantAnnotationRepository>
+{
+    private static final Log LOG = LogFactory.getLog(BaseCachedVariantAnnotationFetcher.class);
+
+    protected String collection;
+    protected VariantAnnotationRepository repository;
+    protected Class<VariantAnnotation> type;
+    protected ExternalResourceFetcher<VariantAnnotation> fetcher;
+    protected ResourceTransformer<VariantAnnotation> transformer;
+    protected Integer maxPageSize;
+
+    public BaseCachedVariantAnnotationFetcher(String collection,
+                                             VariantAnnotationRepository repository,
+                                             Class<VariantAnnotation> type,
+                                             ExternalResourceFetcher<VariantAnnotation> fetcher,
+                                             ResourceTransformer<VariantAnnotation> transformer)
+    {
+        super(collection, repository, type, fetcher, transformer, Integer.MAX_VALUE);
+    }
+
+    public BaseCachedVariantAnnotationFetcher(String collection,
+                                             VariantAnnotationRepository repository,
+                                             Class<VariantAnnotation> type,
+                                             ExternalResourceFetcher<VariantAnnotation> fetcher,
+                                             ResourceTransformer<VariantAnnotation> transformer,
+                                             Integer maxPageSize)
+    {
+        super(collection, repository, type, fetcher, transformer, maxPageSize);
+    }
+   
+    @Override
+    protected String extractId(VariantAnnotation instance)
+    {
+        return instance.getVariantId();
+    }
+
+    @Override
+    protected String extractId(DBObject dbObject)
+    {
+        return (String)dbObject.get("input");
+    }
+
+    @Override 
+    public VariantAnnotation fetchAndCache(String id) throws ResourceMappingException
+    {
+        VariantAnnotation variantAnnotation = null;
+        try {
+            variantAnnotation = super.fetchAndCache(id);
+            if (variantAnnotation == null) {
+                variantAnnotation = new VariantAnnotation(id);
+            } else {
+                variantAnnotation.setSuccessfullyAnnotated(true);
+            }
+        } catch (Exception e) {
+            return new VariantAnnotation(id);
+        }
+        return variantAnnotation;
+    }
+    
+    @Override
+    public List<VariantAnnotation> fetchAndCache(List<String> ids) throws ResourceMappingException
+    {
+        Map<String, VariantAnnotation> variantResponse = this.constructFetchedMap(ids); 
+        for (String variantId : variantResponse.keySet()) {
+            if (variantResponse.get(variantId) == null) {
+                VariantAnnotation variantAnnotation = new VariantAnnotation(variantId);
+                variantResponse.put(variantId, variantAnnotation);
+            } else {
+                variantResponse.get(variantId).setSuccessfullyAnnotated(true);
+            }
+        }
+        // some util function to return response (values) with duplicates
+        // need to pass in a list representing indexes of original request
+        List<VariantAnnotation> values = new ArrayList();
+        for (String id : ids) {
+           values.add(variantResponse.get(id));
+        } 
+        return values;
+    }
+}
--- a/web/src/main/java/org/cbioportal/genome_nexus/web/AnnotationController.java
+++ b/web/src/main/java/org/cbioportal/genome_nexus/web/AnnotationController.java
@@ -235,7 +235,7 @@ public class AnnotationController
     public VariantAnnotation fetchVariantAnnotationByGenomicLocationGET(
         @ApiParam(value="A genomic location. For example 7,140453136,140453136,A,T",
             required = true)
-        @PathVariable @ValidGenomicLocation String genomicLocation,
+        @PathVariable String genomicLocation,
         @ApiParam(value="Isoform override source. For example uniprot",
             required = false)
         @RequestParam(required = false) String isoformOverrideSource,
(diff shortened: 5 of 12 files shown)


ckandoth:
# [Codecov](https://codecov.io/gh/genome-nexus/genome-nexus/pull/371?src=pr&el=h1) Report
> Merging [#371](https://codecov.io/gh/genome-nexus/genome-nexus/pull/371?src=pr&el=desc) into [master](https://codecov.io/gh/genome-nexus/genome-nexus/commit/0f88a535698e431a9427623fe6482cea6886963e&el=desc) will **decrease** coverage by `0.17%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/genome-nexus/genome-nexus/pull/371?src=pr&el=tree)

```diff
@@             Coverage Diff              @@
##             master     #371      +/-   ##
============================================
- Coverage     17.52%   17.34%   -0.18%     
- Complexity      138      139       +1     
============================================
  Files           174      175       +1     
  Lines          2242     2271      +29     
  Branches        287      295       +8     
============================================
+ Hits            393      394       +1     
- Misses         1812     1840      +28     
  Partials         37       37              
```


| [Impacted Files](https://codecov.io/gh/genome-nexus/genome-nexus/pull/371?src=pr&el=tree) | Coverage Δ | Complexity Δ | |
|---|---|---|---|
| [...ice/cached/BaseCachedVariantAnnotationFetcher.java](https://codecov.io/gh/genome-nexus/genome-nexus/pull/371/diff?src=pr&el=tree#diff-c2VydmljZS9zcmMvbWFpbi9qYXZhL29yZy9jYmlvcG9ydGFsL2dlbm9tZV9uZXh1cy9zZXJ2aWNlL2NhY2hlZC9CYXNlQ2FjaGVkVmFyaWFudEFubm90YXRpb25GZXRjaGVyLmphdmE=) | `3.44% <0.00%> (ø)` | `1.00% <0.00%> (?%)` | |

------

[Continue to review full report at Codecov](https://codecov.io/gh/genome-nexus/genome-nexus/pull/371?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/genome-nexus/genome-nexus/pull/371?src=pr&el=footer). Last update [0f88a53...0eb1e91](https://codecov.io/gh/genome-nexus/genome-nexus/pull/371?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

web/src/main/java/org/cbioportal/genome_nexus/web/mixin/VariantAnnotationMixin.java
```diff
@@ -63,6 +63,10 @@ public Integer getStart()
     @ApiModelProperty(value = "List of transcripts", required = false)
     private List<TranscriptConsequence> transcriptConsequences;
 
+    @JsonProperty(value="successfully_annotated", required = true)
+    @ApiModelProperty(value = "Status flag indicating whether variant was succesfully annotated", required = false)
+    private Boolean successfullyAnnotated;
```
inodb: It might be a lot of work to change the model more extensively than adding another field, but some thoughts on this:

- how do we store status per resource? Every resource gets a field `succesfully_annotated`?
- where do we store error messages from e.g. VEP and other resources. Is that going in the response as well? It might be nice if there was some filter to enable/disable verbose error reporting
- how do we store errors in our processing of the the raw resources. E.g. `annotation_summary` related errors because some fields are missing. We wouldn't want the whole request to fail, so some catch all errors solution per variant might be good

Anyway adding the field in this manner can be a good start, but just wanted to mention it

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java
```diff
@@ -322,7 +328,13 @@ else if (ref.length() > 1 || var.length() > 1) {
     public List<String> genomicToEnsemblRestRegion(List<GenomicLocation> genomicLocations)
     {
         List<String> ensemblRestRegionsList = new ArrayList<>();
-        ensemblRestRegionsList.addAll(this.genomicToEnsemblRestRegionMap(genomicLocations).keySet());
+        //ensemblRestRegionsList.addAll(this.genomicToEnsemblRestRegionMap(genomicLocations).keySet());
```
ao508: let's remove this comment

web/src/main/java/org/cbioportal/genome_nexus/web/mixin/VariantAnnotationMixin.java
averyniceday: @inodb good points! Thanks for the feedback. 

- In regards to the first point and third points, what do you mean by resources? Right now we are only adding this successfully annotated flag for variant annotations returned by the annotation controller. The flag just indicates if VEP was able to annotate the passed in variant. I'd be happy to talk about it in more detail to get a more complete picture. 
- This PR is just a first pass for adding the flag. The primary goal was just to ensure we were not silently dropping requested variants. We have discussed expanding it to include error messages but that will have to come down the line. In my picture, we will first need to alter the VEP Wrapper/Runner to include those details in its response.

model/src/main/java/org/cbioportal/genome_nexus/model/VariantAnnotation.java
```diff
@@ -87,6 +88,7 @@ public VariantAnnotation(String variant, String annotationJSON)
         this.variant = variant;
         this.annotationJSON = annotationJSON;
         this.dynamicProps = new LinkedHashMap<>();
+        this.successfullyAnnotated = !(annotationJSON == null);
```
sheridancbio: if we are using a boolean, this is a good name

web/src/main/java/org/cbioportal/genome_nexus/web/mixin/VariantAnnotationMixin.java
averyniceday: also wanted to mention that it seems like the unit/integration tests behave differently depending on the `application.properties` file sitting in the resources directory (e.g test expects Hgvs format but might run region format). I'm planning on creating an issue for ironing that out.

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    protected String extractId(VariantAnnotation instance)
+    {
+        return instance.getVariantId();
+    }
+
+    @Override
+    protected String extractId(DBObject dbObject)
+    {
+        return (String)dbObject.get("input");
```
ao508: minor formatting suggestion : can we add a space between `(String)` and `dbObject`?

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override 
+    public VariantAnnotation fetchAndCache(String id) throws ResourceMappingException
+    {
+        VariantAnnotation variantAnnotation = null;
+        try {
+            variantAnnotation = super.fetchAndCache(id);
+            if (variantAnnotation == null) {
+                variantAnnotation = new VariantAnnotation(id);
+            } else {
+                variantAnnotation.setSuccessfullyAnnotated(true);
```
ao508: so is it necessary to override this value if the model has some logic for resolving the value in the constructor? Or maybe we can move the logic to the getter?

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    public List<VariantAnnotation> fetchAndCache(List<String> ids) throws ResourceMappingException
+    {
+        Map<String, VariantAnnotation> variantResponse = super.constructFetchedMap(ids); 
+        for (String variantId : variantResponse.keySet()) {
+            if (variantResponse.get(variantId) == null) {
+                VariantAnnotation variantAnnotation = new VariantAnnotation(variantId);
+                variantResponse.put(variantId, variantAnnotation);
+            } else {
+                variantResponse.get(variantId).setSuccessfullyAnnotated(true);
```
ao508: same comment here

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java
```diff
@@ -0,0 +1,102 @@
...
+            }
+        }
+        // some util function to return response (values) with duplicates
+        // need to pass in a list representing indexes of original request
+        List<VariantAnnotation> values = new ArrayList();
```
ao508: perhaps to be consistent with other areas of the code we can convert this for-block to its lambda representation? I have no preference either way.

web/src/main/java/org/cbioportal/genome_nexus/web/AnnotationController.java
```diff
@@ -215,7 +215,7 @@ public VariantAnnotation fetchVariantAnnotationGET(
     public VariantAnnotation fetchVariantAnnotationByGenomicLocationGET(
         @ApiParam(value="A genomic location. For example 7,140453136,140453136,A,T",
             required = true)
-        @PathVariable @ValidGenomicLocation String genomicLocation,
+        @PathVariable String genomicLocation,
```
ao508: is this `@ValidGenomicLocation` annotation used anywhere else?

If it's unused anywhere else in the code then perhaps we can remove the class completely

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override 
+    public VariantAnnotation fetchAndCache(String id) throws ResourceMappingException
+    {
+        VariantAnnotation variantAnnotation = null;
+        try {
+            variantAnnotation = super.fetchAndCache(id);
+            if (variantAnnotation == null) {
+                variantAnnotation = new VariantAnnotation(id);
```
sheridancbio: Creates a "failed" VariantAnnotation ?  That is a little bit hidden ... maybe another constructor which takes an explicit second argument for the successfullyAnnotated flag?

web/src/main/java/org/cbioportal/genome_nexus/web/mixin/VariantAnnotationMixin.java
ao508: @inodb One idea we had was to add an additional field that stores the exception/error message - what do you think?

web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
```diff
@@ -67,12 +67,17 @@ public void testAnnotation()
         //////////////////
 
         // for each variant we should have one matching instance, except the invalid one
```
ao508: update this comment to reflect the changes since we no longer remove the invalid variant from the list of responses here

web/src/test/java/org/cbioportal/genome_nexus/web/VEPIdIntegrationTest.java
```diff
@@ -59,7 +59,7 @@ public void testVariantAnnotationById()
 
         // for each dbsnpid we should have one matching variantannotation
         // instance, except the invalid one
-        assertEquals(dbSNPIds.length -1, anns.length);
```
ao508: same comment here - update the comment to reflect the changes made

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java
```diff
@@ -0,0 +1,102 @@
...
+                VariantAnnotation variantAnnotation = new VariantAnnotation(variantId);
+                variantResponse.put(variantId, variantAnnotation);
+            } else {
+                variantResponse.get(variantId).setSuccessfullyAnnotated(true);
+            }
+        }
+        // some util function to return response (values) with duplicates
+        // need to pass in a list representing indexes of original request
+        List<VariantAnnotation> values = new ArrayList();
+        for (String id : ids) {
```
sheridancbio: use the order of the requests to fill the response correctly ... and creates duplicate entries when duplicated in input.
It seems like this is done - maybe the comment above does not need to refer to a util function.

web/src/main/java/org/cbioportal/genome_nexus/web/mixin/VariantAnnotationMixin.java
inodb: @averniceday - gotcha! By resources I meant other annotation sources than vep, e.g. my_variant_info, mutation_assessor etc. One tricky thing with adding fields to the model in this way is that if e.g. we want to track success for my_variant_info and we would follow the same design, we would add it there similarly like: my_variant_info.successfully_annotated. But if we want to reuse the my_variant_info model directly by importing it (they use swagger as well), then we might have to transform their model to include that field, which could be complicated. Using more namespacing/nesting in our response could avoid that, because then a particular subtree of the response could be identical to the raw my_variant_info response. But anyway: one step at a time, just something to be aware of

inodb: @ao508 I think that makes sense. Since we're doing http requests, status codes might be useful as well. But note my point on the nesting. Not sure if that's clear enough I can give an example otherwise

inodb: To elaborate you can do e.g. like this:

```
annotation: {
    vep: {
      successfully_annotated:
      ensembl_transcript_id:
      ... // other vep related fields
    } 
}
```
or try some other namespace options e.g.:
```
annotation: {
   vep: { // vep fields }
   status: {
       vep: {
           http_code: int
           success: bool
           error_message: string
       }
   }
}
```
Advantage of the latter is that if e.g. VEP would have a OpenAPI spec you could import that easily: 
```
annotation: {
    vep: VEPModel // imported from vep
    status: OurStatusModel
}
```
You might be able to achieve the same by extending the VEPModel like:
```
OurVEPModel extends VEPModel {
     // gets all fields from VEPModel through inheritance plus
     successfullyAnnotated: bool
}
```
But that might break certain things downstream that expect the VEPModel as input (e.g. on the frontend or in a custom script somebody was previously using VEP). As well as that there might be conflicts if we add more status related fields that happen to have the same name (unlikely, but still one can avoid it by design). If we want to stick to changing the existing model `VariantAnnotationMixin` i think namespacing it a bit more might help. E.g.

```
OurVEPModel extends VEPModel {
     // gets all fields from VEPModel through inheritance plus
     annotation_source_status: {
         success: bool
         error_message: string
          // etc
     }
}
```
or something along those lines

sheridancbio: I agree that when we modify the model defined by VEP / Ensembl we are blurring an externally defined type (which may evolve as the tool / site is updated). So we are vulnerable to clashes and misinterpretations by clients. It would be safer to keep the original type intact/unmodified and to wrap it inside a larger structure.
Also, each time we change to the model delivered by our API we affect clients, requiring rework. So we will be paying additional development costs by an incremental approach where we add a "successfullyAnnotated" flag now, propagate those changes to clients, and then (in a future revision) add a error_message / failure_description field (maybe also wrapping / relocating the status flag), and then propagate *those* changes to clients.
This is one of the detriments of agile programming - you take small incremental steps, but pay the overhead costs of evolving your way to a future solution which is more general / more perfected. Also, there is no guarantee that you ever get there.
If we all agree that it is better to have an embedded, unmodified version of the VEP response model, then we have to choose between keeping what we have now (a tweaked version) and shipping it before reworking it, or reworking it now (delaying our deployment). I'd be in favor of reworking the model to be an embedded model now, but it does expand the scope of what we have planned for this scrum. Opinions @averyniceday @ao508 @mandawilson ? If we want to recommend the embedded unmodified model, we should talk it through with @n1zea144

inodb: Thanks @sheridancbio! I don't have a strong preference for modifying it now or later.

To add some more context: we have unfortunately already been modifying the VEP model a bit, so maybe it's ok to tweak it further. There are ideas to do a `v2` of this API response. Maybe it's fine that the current one is still a bit hacky as long as we remember the lessons we've learned for the design of a potential new endpoint. Clearly showing in the response what `raw` results are vs `genome nexus` specific is one of those lessons. I have some more ideas about this, but this will be a larger effort.

Good points about the downsides of agile. It often helps you avoid trying to reach perfection, but there can be a lot of overhead with an iterative approach as well.

For this particular PR. I think completely separating the raw response from the status might be a lot of work. A very minimal change though I think is to make the `successfullyAnnotated ` field, a nested object. So instead of:

```
successfully_annotated
```
something like:
```
     annotation_source_status: {
         success: bool
         // TODO: below fields will be implemented later
         // error_message: string
          // etc
     }
```
`annotation_source_status` would then still be a field of the VEPModel but it's nice that all the status related things are grouped already.

That would already be pretty helpful cause you can pass the nested status object around thru various functions more easily on e.g. the frontend and the command line client. When we change that model later on to add more detailed status info (e.g. the error message), then whoever implements those things downstream doesn't have to go looking for which fields are related to status information, because they are all grouped

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java
averyniceday: I think the problem was since we're casting the response to VariantAnnotation type and the response does not naturally have a successfully annotated field, it does not get properly set.

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java
averyniceday: Going to keep it like this (easier for me to read but I don't mind if we change it down the line if someone else feels strongly about it)

web/src/main/java/org/cbioportal/genome_nexus/web/AnnotationController.java
averyniceday: It's not used elsewhere, thinking of keeping it until someone more familiar with the code approves. Seems like this may have been purposely introduced to prevent incorrectly formatted from being submitted (but I guess that is the behavior we are changing)

averyniceday:
@onursumer @leexgh do you guys want to review this before we move forward with this?

web/src/main/java/org/cbioportal/genome_nexus/web/mixin/VariantAnnotationMixin.java
leexgh: > @averniceday - gotcha! By resources I meant other annotation sources than vep, e.g. my_variant_info, mutation_assessor etc. One tricky thing with adding fields to the model in this way is that if e.g. we want to track success for my_variant_info and we would follow the same design, we would add it there similarly like: my_variant_info.successfully_annotated. But if we want to reuse the my_variant_info model directly by importing it (they use swagger as well), then we might have to transform their model to include that field, which could be complicated. Using more namespacing/nesting in our response could avoid that, because then a particular subtree of the response could be identical to the raw my_variant_info response. But anyway: one step at a time, just something to be aware of

@averyniceday i'm working on integrating Oncokb to annotation, in that pr I import their models and use directly, and it might be complicated to change their model to include that field, not sure is there a good way to indicate the status of these added resources?

leexgh: > @averniceday - gotcha! By resources I meant other annotation sources than vep, e.g. my_variant_info, mutation_assessor etc. One tricky thing with adding fields to the model in this way is that if e.g. we want to track success for my_variant_info and we would follow the same design, we would add it there similarly like: my_variant_info.successfully_annotated. But if we want to reuse the my_variant_info model directly by importing it (they use swagger as well), then we might have to transform their model to include that field, which could be complicated. Using more namespacing/nesting in our response could avoid that, because then a particular subtree of the response could be identical to the raw my_variant_info response. But anyway: one step at a time, just something to be aware of

@averyniceday i'm working on integrating Oncokb to annotation, in that pr I import their models and use directly, and it might be complicated to change their model to include that field, not sure is there a good way to indicate the status of these added resources?

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java
```diff
@@ -121,7 +118,7 @@ public T fetchAndCache(String id) throws ResourceMappingException
         }
     }
 
-    public List<T> fetchAndCache(List<String> ids) throws ResourceMappingException
+    public Map<String, T> constructFetchedMap(List<String> ids) throws ResourceMappingException
```
onursumer: We may want to add this method signature to the base interface `CachedExternalResourceFetcher`

web/src/main/java/org/cbioportal/genome_nexus/web/mixin/VariantAnnotationMixin.java
averyniceday: @leexgh I'm not familiar with the OncoKB development, do you mean you are also manually interacting with the VariantAnnotation model to set something related to OncoKB? The new code here does something similar. Let me know if you want to talk about it in more detail!

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java
averyniceday: @onursumer added!

averyniceday: @onursumer added!

averyniceday:
@inodb I've added changes suggested by Onur. Do you think this PR is good to merge? Would like to get it in to prevent it from falling out of date again.

averyniceday:
@inodb I think we'd like to get this in for now and then maybe follow up with a PR in the future. We have already made changes downstream in the annotation pipelines/CVR-fetcher that depend on this change. But I will create a candidate/issue card for this. cc: @sheridancbio @ao508 @mandawilson 

Thanks! Is there a preference for who merges it?

inodb:
@averyniceday sounds good! Feel free to merge

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java
```diff
@@ -0,0 +1,102 @@
...
+        } catch (Exception e) {
+            return new VariantAnnotation(id);
+        }
+        return variantAnnotation;
+    }
+    
+    @Override
+    public List<VariantAnnotation> fetchAndCache(List<String> ids) throws ResourceMappingException
+    {
+        Map<String, VariantAnnotation> variantResponse = super.constructFetchedMap(ids);
```
onursumer: Is there a specific reason to use the `super` keyword here? Why not just `this`?

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java
onursumer: marked as resolved, but I still see the commented out line not removed

web/src/main/java/org/cbioportal/genome_nexus/web/AnnotationController.java
onursumer: `@ValidGenomicLocation` does a pre-validation for this field. So now we want to query the VEP with an invalid genomic location even if we are sure that it's going to fail?

model/src/main/java/org/cbioportal/genome_nexus/model/VariantAnnotation.java
```diff
@@ -302,6 +302,14 @@ public void setOncokbAnnotation(OncokbAnnotation oncokbAnnotation) {
         this.oncokbAnnotation = oncokbAnnotation;
     }
 
+    public void setSuccessfullyAnnotated(Boolean successfullyAnnotated) {
+        this.successfullyAnnotated = successfullyAnnotated;
+    }
+
+    public Boolean getSuccessfullyAnnotated() {
```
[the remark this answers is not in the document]
averyniceday: changed!

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java
averyniceday: changed!

averyniceday:
@onursumer yes, we are purposely removing the pre-validation annotation so the GET will match the POST and respond with a variant (input) and a successfullyAnnotated flag.

PR #376, opened 2020-02-19
git action to test oncokb

leexgh:
the test will run every day at 10am

tried locally and it works:
https://github.com/leexgh/genome-nexus/runs/453855034?check_suite_focus=true


inodb:
@leexgh nice work! 🎉!

PR #386, opened 2020-03-04
add amino acids to annotation summary

inodb:
VEP has a useful field called `amino_acids`, see also:

https://www.genomenexus.org/annotation/7:g.55249071C%3ET?fields=annotation_summary

It's currently not part of annotation_summary. To avoid having to parse ref and alt amino acids on frontend or in the cli this adds the already parsed fields as well, so one gets:

```
aminoAcids: "T/M"
aminoAcidRef: "T"
aminoAcidAlt: "M"
```

Background: these fields are used by the hotspots algorithm, so would be nice to have these in the MAFs annotated by genome nexus


onursumer:
@inodb Test are failing because `AminoAcidsResolver` seems missing in the commit.

inodb:
@onursumer oops - fixed

PR #389, opened 2020-03-17
Add ResponseTimeout Property and Add to Requests to VepWrapper

averyniceday:


service/src/main/java/org/cbioportal/genome_nexus/service/remote/VEPRegionDataFetcher.java
```diff
@@ -27,6 +27,9 @@
     private static final String MAIN_QUERY_PARAM = "variant";
     private static final String PLACEHOLDER = "VARIANT";
 
+    @Value("${gn_vep.timeout.seconds:0}")
```
sheridancbio: I see .. if the property is not specified it defaults to zero .. which means VEP will never internally timeout. That is why the property section is commented out below. Looks fine.

sheridancbio:
I am making a PR now to this fork ... I think it would be good if we do not send the requestTimeout parameter when the value set is zero. By not sending it, we keep the genome-nexus server code compatible with Ensembl and our standalone ec2 instance deployment of the vep service for annotation.genomenexus.org. We would be able to deploy these updates to the current annotation.genomenexus.org server without having to first update VEP by not setting a requestTimeout in that deployment.

PR #392, opened 2020-03-23
Fix #390 KIT annotation issue

### LINKED ISSUE #390: KIT annotation issue
We noticed in the portal on the same protein location, we have different reference AAs.
Here is the link to all KIT samples in MSK portal. [**LINK**](https://cbioportal.mskcc.org/results/mutations?Action=Submit&RPPA_SCORE_THRESHOLD=2.0&Z_SCORE_THRESHOLD=2.0&cancer_study_list=mskimpact&case_set_id=mskimpact_cnaseq&data_priority=0&gene_list=KIT%2520EGFR&geneset_list=%20&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=mskimpact_cna&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=mskimpact_mutations&profileFilter=0&tab_index=tab_visualize)

The issue variants are P573Q and Q573dup. Here are the links to GN with appropriate HGVSg
Q573dup: https://www.genomenexus.org/variant/4:g.55593656_55593657insCAACTTCCTTATGATCACAAATGGGAGTTTCCCAGAAACAGGCTGAGTTTTGGT
P573Q: https://www.genomenexus.org/variant/4:g.55593652C%3EA

The issue was from VEP: https://grch37.rest.ensembl.org/vep/human/hgvs/4:g.55593656_55593657insCAACTTCCTTATGATCACAAATGGGAGTTTCCCAGAAACAGGCTGAGTTTTGGT?content-type=application/json

- [x] Genome Nexus fix
- [x] Genome Nexus Deployed
- [ ] Reannotate mskimpact cases

inodb:
When hgvsp is missing from VEP output and hgvsc contains "dup", use the
entire amino_acids field instead of just the first one. This is in accordance with what CVR is reporting for:

4:g.55593656_55593657insCAACTTCCTTATGATCACAAATGGGAGTTTCCCAGAAACAGGCTGAGTTTTGGT

p.Q575_G592dup

Given the cDNA change it is actually slightly more complicated because it seems to duplicate beyond the exon:

ENST00000288135.5:c.1725_1774+4dup

but maybe `p.Q575_G592dup` is ok

Other changes:
- Unrelated: it seems that we sometimes return hgvspShort without the p.
prefix. I believe that is an error, so fixed those as well.
- Because of reuse of the mock classes, mockito complains about mocking
certain methods that aren't used. In this case I would say reusing the
initialized mocks is useful but not all mock data is useful for testing
all resolvers. So I added the "Silent" runner, we have already been
using this in some places after upgrading to the latest mockito. We
might want to revisit this at a later point, but seemed out of the scope
for this PR

### DIFF
--- a/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/ProteinChangeResolver.java
+++ b/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/ProteinChangeResolver.java
@@ -133,7 +133,7 @@ public class ProteinChangeResolver
                 String variantClassification = this.variantClassificationResolver.resolve(
                     variantAnnotation, transcriptConsequence);
 
-                hgvspShort = "*" + String.valueOf(pPos);
+                hgvspShort = "p.*" + String.valueOf(pPos);
 
                 if (variantClassification != null && variantClassification.toLowerCase().startsWith("frame_shift")) {
                     hgvspShort += "fs*";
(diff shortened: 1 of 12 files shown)


zhx828:
Thanks for the quick fix.

PR #393, opened 2020-03-24
Fix #387 Add nucelotide context

### LINKED ISSUE #387: Show bases in a window around a mutation
For computing mutational signatures using SNVs one needs the bases 5' and 3' around the position of the mutation. Similarly for the hotspot algorithm. It would be nice if genome nexus could provide these

inodb:
Add another annotation service called nucleotide context that uses
ensembl to pull one base 5' and 3' around a SNV. This is useful for
calculating e.g. signatures and hotspot analyses.


inodb:
@onursumer @leexgh not sure what's going on with test reporting of Travis CI in GitHub but it passes: https://travis-ci.org/github/genome-nexus/genome-nexus/builds/666565484

Issue #403, opened 2020-04-15
nucleotide_context not returned for POST of genomic location


inodb:
Try e.g.

```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' -d '[{"chromosome":"7","start":140453136,"end":140453136,"referenceAllele":"A","variantAllele":"T"}]' 'https://www.genomenexus.org/annotation/genomic?fields=hotspots%2Cmutation_assessor%2Cnucelotide_context'  | python -m json.tool | less -S
```

and search for `nucleotide_context`. Only works for GET requests atm

inodb:
Turned out there was a typo: `nucelotide_context` 🤦 . We should prolly handle incorrect field names in some way

PR #404, opened 2020-04-16
Add integration tests for genomic location

inodb:
Added some testing for the genomic location endpoint While trying to
address #403, which turned out to be a typo by the API consumer (i.e.
yours truly).

### DIFF
--- a/web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
+++ b/web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
@@ -1,4 +1,6 @@
 package org.cbioportal.genome_nexus.web;
+
+import org.cbioportal.genome_nexus.model.GenomicLocation;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.boot.json.JsonParser;
@@ -27,6 +29,10 @@ public class AnnotationIntegrationTest
 {
     private final static String BASE_URL = "http://localhost:38896/annotation/";
 
+    private final static String BASE_URL_GENOMIC_LOCATION = "http://localhost:38896/annotation/genomic/";
+
+    private final static String ALL_ENSEMBL_FIELDS = "fields=annotation_summary,nucleotide_context";
+
     private RestTemplate restTemplate = new RestTemplate();
 
     private Map<String, Object> fetchVariantAnnotationGET(String variant)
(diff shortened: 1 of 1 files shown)


web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
```diff
@@ -136,4 +155,86 @@ public void testLongDeletion()
...
+    public void testGenomicLocationSNVGET() {
+        String genomicLocation = "7,140453136,140453136,A,T";
+
+        Map<String, Object> response = this.fetchVariantAnnotationByGenomicLocationGET(genomicLocation);
+
+        // TODO: somehow only the POST contains hgvsg but not GET We should
+        // maybe do a more general test that checks for object equality between
+        // GET response and POST response for all types of endpoints. This is
+        // not enforced through typing because a lot of the fields are optional.
+        // assertEquals(response.get("hgvsg"), "7:g.140453136A>T");
```
onursumer, line 169: Interesting, do those fields exist for the original web service request?

web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
```diff
@@ -136,4 +155,86 @@ public void testLongDeletion()
...
+        String seq = (String) annotation.get("seq");
+        assertEquals("CAC", seq);
+    }
+
+    @Test
+    public void testGenomicLocationNucleotideContextPOST() {
+        String genomicLocation = "7,140453136,140453136,A,T";
+        GenomicLocation[] genomicLocations = {
+            new GenomicLocation() {{
+                setChromosome(genomicLocation.split(",")[0]);
```
onursumer, line 226: maybe do `String[] parts = genomicLocation.split(",")` above to avoid repetitive split calls, not a huge deal though.

inodb: yeah there should prolly be some util function to convert genomicLocationString -> GenomicLocation

web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
inodb: Doesn't look like it: http://grch37.rest.ensembl.org/vep/human/hgvs/7:g.140453136A%3ET?content-type=application/json&hgvs=1&hgvsg=1. I guess we are adding them in the POST requests somewhere

inodb:
@leexgh seems fixed now. I think there are some occasional intermittent failures of the integration tests because of how there are multiple tests apps started at the same time running on different ports. Maybe they are interfering with each other

PR #407, opened 2020-04-30
Add a signal annotation field for annotation endpoint

### LINKED ISSUE #406: Enrich variant annotation with Signal info
- Add support for a new field option named `signal` for the annotation endpoint
- Add a `SignalEnricher` to match a given variant annotation to a signal mutation by genomic location
- Attach the enrichment as a new field to the variant annotation, similar to the other enrichments we have

onursumer:
Fix #406

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/SignalAnnotation.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/SignalAnnotation.java
@@ -1,0 +1,37 @@
+package org.cbioportal.genome_nexus.model;
+
+
+import java.util.Collections;
+import java.util.List;
+
+public class SignalAnnotation
+{
+    private String license;
+
+    private List<SignalMutation> annotation;
+
+    public SignalAnnotation()
+    {
+        this.license = "https://www.signaldb.org/about";
+        this.annotation = Collections.emptyList();
+    }
+
+    public String getLicense()
+    {
+        return license;
+    }
+
+    public void setLicense(String license)
+    {
+        this.license = license;
+    }
+    public List<SignalMutation> getAnnotation()
+    {
+        return annotation;
+    }
+
+    public void setAnnotation(List<SignalMutation> annotation)
+    {
+        this.annotation = annotation;
+    }
+}
--- a/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/SignalMutationRepository.java
+++ b/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/SignalMutationRepository.java
@@ -9,4 +9,11 @@ public interface SignalMutationRepository extends MongoRepository<SignalMutation
 {
     List<SignalMutation> findByHugoGeneSymbol(String hugoGeneSymbol);
     List<SignalMutation> findByHugoGeneSymbolIn(List<String> hugoGeneSymbol);
+    List<SignalMutation> findByChromosomeAndStartPositionAndEndPositionAndReferenceAlleleAndVariantAllele(
+        String chromosome,
+        Long startPosition,
+        Long endPosition,
+        String referenceAllele,
+        String variantAllele
+    );
 }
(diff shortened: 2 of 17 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/enricher/SignalAnnotationEnricher.java
```diff
@@ -0,0 +1,31 @@
...
+    private final SignalMutationService signalMutationService;
+
+    public SignalAnnotationEnricher(SignalMutationService signalMutationService) {
+        this.signalMutationService = signalMutationService;
+    }
+
+    @Override
+    public void enrich(VariantAnnotation annotation) {
+        SignalAnnotation signalAnnotation = new SignalAnnotation();
+        String hgvsg = annotation.getHgvsg();
```
onursumer, line 22: Always using `hgvgs` even if the actual query is originated from a genomic location query. It might be better to use the original query parameters if available, but we are only passing the `annotation` as a parameter here. May need to improve the enricher interface for that.

model/src/main/java/org/cbioportal/genome_nexus/model/SignalAnnotation.java
```diff
@@ -0,0 +1,37 @@
...
+
+
+import java.util.Collections;
+import java.util.List;
+
+public class SignalAnnotation
+{
+    private String license;
+
+    private List<SignalMutation> annotation;
```
onursumer, line 11: `SignalMutation` has a lot of data, not sure if we need all of it. We may want to return a partial response here instead.

service/src/main/java/org/cbioportal/genome_nexus/service/enricher/SignalAnnotationEnricher.java
inodb: agreed, might be good to improve the enricher interface

inodb: maybe good to open a separate issue?

model/src/main/java/org/cbioportal/genome_nexus/model/SignalAnnotation.java
inodb: i see, yeah maybe only the fields that we are showing in signaldb.org? 

E.g. penetrance, %, % biallelic, % somatic driver, % germline pathogenic and broken down by cancer type

inodb: im not sure how much more data there is in signalmutation, so if it's not too crazy, maybe ok. Could you add some example response to the description of the PR + maybe in a test?

inodb:
Thanks @onursumer - overall this looks good to me! Made a few comments

service/src/main/java/org/cbioportal/genome_nexus/service/enricher/SignalAnnotationEnricher.java
onursumer: open a new issue #411

onursumer:
Found a bug (#410) while testing the enricher. We can merge this PR but annotations with insertion or deletion won't get enriched until we fix it.

persistence/src/main/java/org/cbioportal/genome_nexus/persistence/SignalMutationRepository.java
```diff
@@ -9,4 +9,11 @@
 {
     List<SignalMutation> findByHugoGeneSymbol(String hugoGeneSymbol);
     List<SignalMutation> findByHugoGeneSymbolIn(List<String> hugoGeneSymbol);
+    List<SignalMutation> findByChromosomeAndStartPositionAndEndPositionAndReferenceAlleleAndVariantAllele(
```
inodb, line 12: @onursumer this name is a little funky but I guess it gets the point across :). Maybe something like getByGenomicLocationFields or Properties?

onursumer: @inodb the name has to be exactly like this to make it work automagically :) Spring Boot translates this into a proper DB query.

PR #412, opened 2020-05-12
Fix fromHgvs method for insertion and deletion types

### LINKED ISSUE #410: GenomicVariantUtil.fromHgvs does not work as expected for insertions and deletions
An insertion in hgvs notation like `13:g.32906804_32906805insTTAG` is converted as `13,32906804,32906805,,TTAG` but it should actually be `13,32906804,32906805,-,TTAG`.

Similarly, a deletion like `13:g.32893385_32893389delTATTC` is converted as `13,32893385,32893389,TATTC,XXXXX`, but it should be `13,32893385,32893389,TATTC,-`.

onursumer:
Fix #410


component/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -248,15 +248,16 @@ private static String getRefFromHgvs(String hgvs, GenomicVariant.Type type) {
                 }
                 return s.trim();
             case INSERTION:
-                return "";
+                return "-";
```
onursumer, line 251: Simply replaced empty strings with `-`, not sure if this would break something else

component/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -248,15 +248,16 @@ private static String getRefFromHgvs(String hgvs, GenomicVariant.Type type) {
...
             case DELETION:
                 s = getPattern("(?<=[a-z+>])[ATCG]*$", hgvs);
                 if (s == null || s.trim().length() == 0) {
-                    return "";
+                    return "-";
                 }
                 return s.trim();
             case INDEL:
-                // use a string of "X" equal to length of alt allele TODO : check if this is used anywhere and drop this if not .. also check if this is needed above for DELETION
+                // use a string of "X" equal to length of alt allele
```
onursumer, line 259: leaving this as is since not sure if we are using this anywhere

component/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -272,13 +273,11 @@ private static String getAltFromHgvs(String hgvs, GenomicVariant.Type type) {
...
                 if (s == null || s.trim().length() == 0) {
-                    return "";
+                    return "-";
                 }
                 return s.trim();
             case DELETION:
-                // TODO : here, the deleted nucleotids are optional .. but we convert them into a string of "X" if they are present. Is this correct?
-                s = getPattern("(?<=[a-z+>])[ATCG]*$", hgvs);
-                return sameLengthUnknownNucleotideString(s);
+                return "-";
```
onursumer, line 280: simply returning `-` for deletions instead of same length `X`s. I think we don't need `X`s here

component/src/test/java/org/cbioportal/genome_nexus/util/GenomicVariantUtilTest.java
```diff
@@ -0,0 +1,68 @@
+package org.cbioportal.genome_nexus.util;
+
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class GenomicVariantUtilTest {
```
onursumer, line 8: Apparently we have test cases for `GenomicVariantUtil`, but in the wrong module with a wrong name... https://github.com/genome-nexus/genome-nexus/blob/b3bab9c9532e1641bbf0440bd992e98a291951ce/service/src/test/java/org/cbioportal/genome_nexus/util/GenomicVariantTest.java#L9

We should probably move those tests here as well, and remove the duplicate ones that I just added.

onursumer:
@leexgh good point. Not sure if it is supported but there is a `TODO` mentioning `dup`: https://github.com/genome-nexus/genome-nexus/blob/98ba43a8bb4b269262e662726e4a45bff83fb6d8/component/src/test/java/org/cbioportal/genome_nexus/util/GenomicVariantUtilTest.java#L141

sheridancbio:
@leexgh @onursumer I think that dup is not supported when the inputs come from records encoded in GenomicVariant models (like from a MAF file). If you directly put in HGVS references into the annotation api, that may work / be passed along to VEP when using the ensembl annotation link. For local VEP wrapper we convert first to GenomicVariant and then to Region format, so I think dup would not work in that case.

component/src/test/java/org/cbioportal/genome_nexus/util/GenomicVariantUtilTest.java
sheridancbio: 👍

PR #416, opened 2020-06-03
Enable ACCEPT_SINGLE_VALUE_AS_ARRAY only for MyVariantInfoFetcher

### LINKED ISSUE #415: MongoDB entity mapper is broken due to a global deserialization setting
https://github.com/genome-nexus/genome-nexus/blob/afd0f89c984c39367ff6e423de95c2b65524f831/service/src/main/java/org/cbioportal/genome_nexus/service/config/ExternalResourceObjectMapper.java#L27

This allows parsing a single value as a list when reading from a web service. But MongoDB entity mapper cannot handle this since we cache the original response as is.

Short term solution is to disable this global feature and manually normalize data before saving to DB.

For a long term solution we should investigate if it is possible to use the same object mapper for MongoDB entities.

onursumer:
Fix #415


inodb:
@onursumer when releasing this, should we be clearing the myvariant.info cache?

onursumer:
> @onursumer when releasing this, should we be clearing the myvariant.info cache?

@inodb I think we should. the cache is mostly likely already compromised.

PR #429, opened 2020-07-14
Retain original query input in annotation JSON response

ao508:
This commit addresses an issue where the original input query did not
match the 'input' field in the JSON annotation response due to a shift in
the value of 'start_position' for certain types of genomic variants when
converting genomic locations to HGVSg format strings.

As a result some variants submitted for annotation through the Genome
Nexus annotation pipeline were considered failed annotations since the
input query did not match the 'variant' field exactly in the JSON
response which is how the client matches the queries back to the records
submitted for annotation.

Signed-off-by: Angelica Ochoa <15623749+ao508@users.noreply.github.com>

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/GenomicLocation.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/GenomicLocation.java
@@ -1,5 +1,7 @@
 package org.cbioportal.genome_nexus.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
 public class GenomicLocation
 {
     private String chromosome;
@@ -7,6 +9,8 @@ public class GenomicLocation
     private Integer end;
     private String referenceAllele;
     private String variantAllele;
+    @JsonIgnore
+    private String originalInput;
 
     public String getChromosome() {
         return chromosome;
@@ -46,6 +50,14 @@ public class GenomicLocation
 
     public void setVariantAllele(String variantAllele) {
         this.variantAllele = variantAllele;
+    }
+
+    public String getOriginalInput() {
+        return originalInput;
+    }
+
+    public void setOriginalInput(String originalInput) {
+        this.originalInput = originalInput;
     }
 
     public String toString() {
--- a/web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
+++ b/web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
@@ -270,6 +270,18 @@ public class AnnotationIntegrationTest
         assertEquals("CAC", seq);
     }
 
+    @Test
+    public void testVariantAnnotationOriginalQuery() {
+        String expectedConvertedVariant = "4:g.55152096_55152107delTCATGCATGATT";
+        String genomicLocationString = "4,55152095,55152107,ATCATGCATGATT,A";
+        GenomicLocation[] genomicLocations = {
+            genomicLocationStringToGenomicLocation(genomicLocationString)
+        };
+        List<Map<String, Object>> response = this.fetchVariantAnnotationByGenomicLocationPOST(genomicLocations);
+        assertEquals(genomicLocationString, response.get(0).get("originalVariantQuery"));
+        assertEquals(expectedConvertedVariant, response.get(0).get("variant"));
+    }
+
     private GenomicLocation genomicLocationStringToGenomicLocation(String genomicLocation) {
         return new GenomicLocation() {{
             setChromosome(genomicLocation.split(",")[0]);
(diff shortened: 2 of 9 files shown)


model/src/main/java/org/cbioportal/genome_nexus/model/GenomicLocation.java
```diff
@@ -7,6 +7,7 @@
     private Integer end;
     private String referenceAllele;
     private String variantAllele;
+    private String originalInput;
```
onursumer, line 10: Not sure if `originalInput` should be part of `GenomicLocation` model. Technically input can be anything, right? Might be better to have a mapping somewhere else instead of tying the input to the model.

ao508: So the POST accepts a list of `GenomicLocation` objects.. this was my attempt to retain the original input as provided to the POST method since in the service layer the `GenomicLocation` objects are converted into their HGVSg format representations and the `variantAnnotation` services basically loses knowledge of what the original input might have been in the GenomicLocation objects.

Do you have suggestions for how this might be implemented better? @onursumer

onursumer: We can probably add a separate POST which accepts a list of strings instead of genomic location objects. With this change, will we always have to provide the original input value for POST queries from now on?

ao508: @onursumer So the string representations of the `GenomicLocation` objects? Yeah, I think that would work.. will test first of course but I want to say yes it would.

ao508: actually I think that some additional changes would be required as well to prevent the normalization of the input genomic location string

ao508: so we may still run into the case where the start position provided in the input does not match the start position in the annotation json response

ao508: @onursumer I confirmed that the same issue persists. To make it work this way we would need to skip the step where the genomic location is normalized by the NotationConverter. This is the step that is causing the mismatch between the start positions of the input query and the returned annotation JSON. If there is a common prefix in the reference allele and the variant allele then it is removed and the start position is adjusted accordingly.

onursumer: I guess we can keep it this way if it is too much work to separate the input field from the model

web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
```diff
@@ -270,13 +270,26 @@ public void testGenomicLocationNucleotideContextPOST() {
...
+    public void testVariantAnnotationOriginalQuery() {
+        String expectedConvertedVariant = "4:g.55152096_55152107delTCATGCATGATT";
+        String genomicLocationString = "4,55152095,55152107,ATCATGCATGATT,A";
+        GenomicLocation[] genomicLocations = {
+            genomicLocationStringToGenomicLocation(genomicLocationString)
+        };
+        List<Map<String, Object>> response = this.fetchVariantAnnotationByGenomicLocationPOST(genomicLocations);
+        assertEquals(genomicLocationString, response.get(0).get("originalVariantQuery"));
+        assertEquals(expectedConvertedVariant, response.get(0).get("variant"));
+    }
```
inodb, line 283: is `fetchVariantAnnotationByGenomicLocationPOST` the only endpoint where we will have `originalVariantQuery`?

It seems like we might need this anywhere where we are doing a POST regardless of input being genomic location or hgvs?

That said I think this if fine for a quick fix

ao508:
@inodb @onursumer Made a small tweak to hide the `originalInput` field for `GenomicLocation` on the swagger page. PR is rebased and @sheridancbio is updating the `genome-nexus-java-api-client`, will share the PR to that soon.

web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
```diff
@@ -260,7 +260,7 @@ public void testGenomicLocationNucleotideContextGET() {
     public void testGenomicLocationNucleotideContextPOST() {
         String genomicLocation = "7,140453136,140453136,A,T";
         GenomicLocation[] genomicLocations = {
-            genomicLocationStringToGenomicLocation(genomicLocation)
+	    genomicLocationStringToGenomicLocation(genomicLocation)
```
leexgh, line 263: Good to add indentation here

ao508:
@inodb @onursumer Anything else to add on this or is it good to go?

onursumer:
@ao508 I see a couple of minor styling/indentation comments not resolved yet, other than that I think it is good to go.

ao508:
@onursumer Fixed the whitespace

Issue #428, opened 2020-07-10
Leave hgvspShort notation empty for mutations in UTR regions (*123*)


leexgh:
Some silent mutations' hgvspShort is not correct. See https://github.com/genome-nexus/genome-nexus/issues/427 
We annotate silent mutations' hgvspShort as `*123*`, probably should leave those hgvspShort empty. 
This is part of epic https://github.com/cBioPortal/cbioportal/issues/7479

PR #430, opened 2020-07-15
remove hgvspShort for mutations in UTR regions

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/428


inodb:
@leexgh Nice!

leexgh:
@inodb not sure if I should remove these https://github.com/genome-nexus/genome-nexus/pull/430/files#diff-979e975054eb3f8fc83702a7c7104c7cL123-L145

PR #431, opened 2020-07-22
don't pull mutation assessor by default

leexgh:


web/src/main/java/org/cbioportal/genome_nexus/web/AnnotationController.java
```diff
@@ -107,7 +107,7 @@ public AnnotationController(VariantAnnotationService hgvsVariantAnnotationServic
             String token,
         @RequestParam(required = false)
         @ApiParam(value="Comma separated list of fields to include (case-sensitive!). " +
-            "For example: hotspots,mutation_assessor", required = false, defaultValue = "hotspots,mutation_assessor")
+            "For example: hotspots", required = false, defaultValue = "hotspots")
```
inodb, line 110: @leexgh would be good to list all the options at some point

PR #456, opened 2020-10-13
Add search endpoint for signal mutations

onursumer:
Supports search by gene, hgvs variant, and region.

Related to knowledgesystems/signal#65

### DIFF
--- a/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/SignalMutationRepository.java
+++ b/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/SignalMutationRepository.java
@@ -1,6 +1,7 @@
 package org.cbioportal.genome_nexus.persistence;
 
 import org.cbioportal.genome_nexus.model.SignalMutation;
+import org.springframework.cache.annotation.Cacheable;
 import org.springframework.data.mongodb.repository.MongoRepository;
 
 import java.util.List;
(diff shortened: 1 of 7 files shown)


persistence/src/main/java/org/cbioportal/genome_nexus/persistence/SignalMutationRepository.java
```diff
@@ -1,12 +1,16 @@
...
 import org.cbioportal.genome_nexus.model.SignalMutation;
+import org.springframework.cache.annotation.Cacheable;
 import org.springframework.data.mongodb.repository.MongoRepository;
 
 import java.util.List;
 
 public interface SignalMutationRepository extends MongoRepository<SignalMutation, String>
 {
+    @Cacheable("allSignalMutations")
+    List<SignalMutation> findAll();
```
onursumer, line 12: This keeps all the mutations in memory. If this is not desired, we can process the mutations and only keep the search fields in another partial model. Alternatively, we can perform custom DB queries instead of pulling all mutations at once.

inodb:
@onursumer deployed here: https://beta.genomenexus.org/signal/search/BRCA1?limit=1. See commit for more notes: https://github.com/knowledgesystems/knowledgesystems-k8s-deployment/commit/09b348ee3c592167aabe3d71e82a1abf97c334eb. I had to mess with config a little bit to support the increased mem usage and the caching that happens at time of first request. Maybe ok for now. I can change the readiness to use the signal query endpoint so it's fully cached before users are able to access it. Let's monitor mem usage a bit

inodb (approved):
With the changes there's a short bump in memory usage now on init (since that's when it's building the index for signal). It might be nice to log some info (e.g. "start indexing signal", "finished indexing signal"), but LGTM

Screen Shot 2020-10-26 at 1 17 35 PM

PR #457, opened 2020-10-20
MyVariantInfo annotation bug fix

ao508:
The map constructed during POST requests was using the `query` field instead of `variant` field which is what should have been used to match with the appropriate `MyVariantInfo` objects and `VariantAnnotation` objects.

Signed-off-by: Angelica Ochoa <15623749+ao508@users.noreply.github.com>

Co-authored-by: Avery Wang <18199796+averyniceday@users.noreply.github.com>
Co-authored-by: Robert Sheridan <7747489+sheridancbio@users.noreply.github.com>
Co-authored-by: Manda Wilson <1458628+mandawilson@users.noreply.github.com>

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/VariantAnnotation.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/VariantAnnotation.java
@@ -263,6 +263,7 @@ public class VariantAnnotation
         this.mutationAssessorAnnotation = mutationAssessorAnnotation;
     }
 
+    @Field(value="my_variant_info")
     public MyVariantInfoAnnotation getMyVariantInfoAnnotation() {
         return myVariantInfoAnnotation;
     }
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/VariantAnnotation.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/VariantAnnotation.java
@@ -263,7 +263,6 @@ public class VariantAnnotation
         this.mutationAssessorAnnotation = mutationAssessorAnnotation;
     }
 
-    @Field(value="my_variant_info")
     public MyVariantInfoAnnotation getMyVariantInfoAnnotation() {
         return myVariantInfoAnnotation;
     }
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/internal/BaseVariantAnnotationServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/internal/BaseVariantAnnotationServiceImpl.java
@@ -229,7 +229,7 @@ public abstract class BaseVariantAnnotationServiceImpl implements VariantAnnotat
                 postEnrichmentService.enrichAnnotations(variantAnnotations);
             }
         } catch (VariantAnnotationWebServiceException e) {
-            LOG.warn(e.getLocalizedMessage());
+            LOG.warn(e.getLocalizedMessage(), e);
         }
 
         return variantAnnotations;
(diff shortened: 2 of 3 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/internal/BaseVariantAnnotationServiceImpl.java
```diff
@@ -229,7 +229,7 @@ private VariantAnnotation getVariantAnnotation(String variant, EnrichmentService
                 postEnrichmentService.enrichAnnotations(variantAnnotations);
             }
         } catch (VariantAnnotationWebServiceException e) {
-            LOG.warn(e.getLocalizedMessage());
+            LOG.warn(e.getLocalizedMessage(), e);
```
ao508, line 232: added this here since it was very useful in tracking down the error causing the issue fixed by this pr

sheridancbio (approved):
👍

thomasyu888:
Are there fields in the "query" field that we will be missing by making this change?

ao508:
@thomasyu888 "query" was just used for this map construction so nothing should be affected by not using it anymore in this map

model/src/main/java/org/cbioportal/genome_nexus/model/VariantAnnotation.java
```diff
@@ -263,6 +263,7 @@ public void setMutationAssessorAnnotation(MutationAssessorAnnotation mutationAss
         this.mutationAssessorAnnotation = mutationAssessorAnnotation;
     }
 
+    @Field(value="my_variant_info")
```
onursumer, line 266: Not sure if we need this one, since `my_variant_info` field should not exist in the `vep.annotation` collection.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/MyVariantInfoServiceImpl.java
```diff
@@ -65,7 +65,7 @@ public MyVariantInfo getMyVariantInfoByHgvsVariant(String variant)
             myVariantInfos = this.getMyVariantInfoByMyVariantInfoVariant(queryVariants);
             // manually set the original hgvs variant field
             for (MyVariantInfo myVariantInfo: myVariantInfos) {
-                myVariantInfo.setHgvs(queryToVariant.get(myVariantInfo.getQuery()));
+                myVariantInfo.setHgvs(queryToVariant.get(myVariantInfo.getVariant()));
```
onursumer, line 68: We need to make sure that My Variant Info returns a valid `variant` for all queries. As far as I remember query is our original input, variant is something returned by MVI.

ao508: @onursumer the query field was actually null in most of the cases we were annotating which is why we weren’t getting all of the MVI annotations back for the variants we submitted in POST requests

sheridancbio: @onursumer maybe there is some bug which is causing the getQuery() field to be blanked out? If you can find and fix such a bug so that getQuery() will not return null for certain cases (perhaps occurring when a GET() request to annotation/genomic is first hit for a variant and then a POST() request follow subsequently) maybe fixing that bug would make this change unnecessary. (and maybe fix other problems we have not yet noticed) cc: @n1zea144

onursumer: @sheridancbio @ao508 I don't see `setQuery()` being manually called anywhere in our codebase. I think `query` is/was part of the My Variant Info response. Seems like at some point their model changed, so that we started to get `null` for this field for any variant that hasn't been already cached.

These types of bugs are difficult to detect unless we constantly monitor changes in external models. As long as we rely on live web services which frequently update their model, unfortunately we will continue having these types of issues.

sheridancbio:
I believe this PR can be closed. A new patch is in place now for our internal importers which removes the clinvar field from myvariant.info responses. Correcting the model has been made an issue here : #460 

I am closing this PR, but if anyone wants it to be reopened in order to use changes while fixing #460 or related problems, feel free.

PR #462, opened 2020-10-28
Enable search by alteration for signal mutations

onursumer:
Related to knowledgesystems/signal#65

- Enable basic search by alteration (e.g. `V600`)
- Break down `SignalQueryServiceImpl` into multiple components.


inodb (approved):
Awesome 🥇

Issue #465, opened 2020-11-16
Some del and Indel queries are not working for signal and my variant info


leexgh:
indel and del sometimes not working for signal and my variant info, because myvariant info are using [this format](https://varnomen.hgvs.org/recommendations/DNA/) of hgvs (My variant info doc: https://docs.myvariant.info/en/latest/doc/data.html?highlight=deletion%20#id-field).
Signal is using genome nexus generated hgvs:
https://github.com/genome-nexus/genome-nexus/blob/master/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java#L140

Other things need to investigate:
- all indel are not working for signal, even with the NotationConverter format of hgvs
- POST hgvs doesn't work for my variant info, but GET hgvs works
 
See details in table:
https://docs.google.com/spreadsheets/d/1cmOIIn2ZbH8pU5boZVB6UkjuzZu_zmyyWbFonBByajI/edit#gid=0

Issue #460, opened 2020-10-27
myvariant.info enrichment breaking due to clinvar response model


sheridancbio:
Requests sent to myvariant.info generate responses which are sometimes failing to be fully parsed. It appears that some expected subelements in the clinvar element are modeled in a way which sometimes differs from the observed responses from genome nexus.

One observed case is MyVariantInfo.clinvar.hgvs.coding, which sometimes contains an array of string in the response but for other queries contains a simple string:

**coding as a string**

```
"clinvar": {
    "_license": "http://bit.ly/2SQdcI0",
    "allele_id": 691874,
    "alt": "C",
    "chrom": "5",
    "cytogenic": "5q11.2",
    "gene": {
      "id": "4214",
      "symbol": "MAP3K1"
    },
    "hg19": {
      "end": 56178615,
      "start": 56178615
    },
    "hg38": {
      "end": 56882788,
      "start": 56882788
    },
    "hgvs": {
      "coding": "NM_005921.2:c.3588A>C",
      "genomic": [
        "NC_000005.10:g.56882788A>C",
        "NC_000005.9:g.56178615A>C",
        "NG_031884.1:g.72716A>C"
      ]
    },
```
**coding as an array**

```
"clinvar": {
    "_license": "http://bit.ly/2SQdcI0",
    "allele_id": 138281,
    "alt": "T",
    "chrom": "1",
    "cytogenic": "1p31.3",
    "gene": {
      "id": "3716",
      "symbol": "JAK1"
    },
    "hg19": {
      "end": 65312368,
      "start": 65312368
    },
    "hg38": {
      "end": 64846685,
      "start": 64846685
    },
    "hgvs": {
      "coding": [
        "LRG_1398t1:c.1951G>A",
        "NM_001320923.1:c.1951G>A",
        "NM_001321852.2:c.1951G>A",
        "NM_001321853.2:c.1951G>A",
        "NM_001321854.2:c.1951G>A",
        "NM_001321855.2:c.1951G>A",
        "NM_001321856.1:c.1951G>A",
        "NM_001321857.2:c.1948G>A",
        "NM_002227.4:c.1951G>A"
      ],

```  

Another subelement which may differ is the rcv subelement (which is modeled as a list of rcv objects which contain 4 simple string members) However, the response from the myvariant.info server violates these expectations for some queries, including some from the integration tests. Possibly this error is that the rcv subelement is now optional (not always provided)

**error reported during run of integration tests**

[...]

Issue #276, opened 2019-05-29
Handle NA reference_allele for some INS


leexgh:
Reference_allele should be `-` but in many studies they are `NA` or multiple `-`.

jjgao:
I guess it doesn't hurt to do this but the data should and will be fixed: https://github.com/cBioPortal/datahub/issues/621

leexgh:
@jjgao That would be great! Thanks! But maybe I can still do it for now, since it doesn't hurt and not so complicated

PR #466, opened 2020-11-17
fix del and indel format queries

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/465

**Description**
Signaldb needs my variant info and signal annotation data. For `del` and `indel`, the official hgvsg format is https://varnomen.hgvs.org/recommendations/DNA/

Correct format:
- 13:g.32914438del (13,32914438,32914438,T,-)
- 17:g.41276045_41276046del (17,41276045,41276046,CT,-)
- 17:g.7579363delinsTTT (17,7579363,7579363,A,TTT)
- 7:g.55242467_55242486delinsTC (7,55242467,55242486,AATTAAGAGAAGCAACATCT,TC)

Hgvsg in genome nexus:
- 13:g.32914438_32914438delT (13,32914438,32914438,T,-)
- 17:g.41276045_41276046delCT (17,41276045,41276046,CT,-)
- 17:g.7579363delAinsTTT (17,7579363,7579363,A,TTT)
- 7:g.55242467_55242486delAATTAAGAGAAGCAACATCTinsTC (7,55242467,55242486,AATTAAGAGAAGCAACATCT,TC)

My variant info supports the official format, while signal is using the genome nexus format. So for a `del` or `indel`, it can't return both sources when giving input in one of the formats.

**Solution**
In this pr I convert hgvsg to the official format. So some format converter for genomic location and hgvsg are affected.

**Expected results**
GET /annotation/{variant} endpoint should be able to handle the correct format, and be able to get both signal and my_variant_info in return.

**Other issues**
1. My variant info issue is still not fixed (https://github.com/genome-nexus/genome-nexus/issues/460). I encounter the same issue when I test it locally(e.g. 7:g.55242467_55242486delinsTC), but for the same variant I can get results from production. Probably merging this pr will still be able to get results. **Good to try with beta instance first.**

2. https://github.com/genome-nexus/genome-nexus/issues/467


leexgh:
@inodb @onursumer @SRodenburg To be honest I'm not so confident on this pr, not sure if it will break anything. Could you please review it carefully?

leexgh:
Another solution is we can create a converter just for signal, so it won't break other things, but will introduce more duplicated converters

inodb (approved):
Thanks @leexgh ! This looks good to me

I think in general we are relying a lot on hgvs as is clear from how much code this PR touches.  I think in most cases we should use genomicLocation as the key or some other unique identifier. Note also that it might for instance be possible that one annotation source uses delACTinsT and another resource use delinsT. Using some single variant object where one can request multiple representations of the same object might be a useful extension. Similarly on the variant page of genome nexus we might want to show common alternative "spellings" of hgvsg

onursumer (commented):
Thanks for updating all the mock data with the correct hgvsg format 🙏

sheridancbio:
A few weeks ago the backend team worked on a tool to overlay gnomAD allele frequency annotation from myvariant.info onto a MAF variant file via a standalone python script. The logic from the genomenexus NotationConverter was involved and we noticed the difference in hgvs format addressed in this PR. The documentation for the Ensembl VEP webAPI shows the "fully-specified" format for deletions or indels : https://rest.ensembl.org/documentation/info/vep_hgvs_get  .. in here the query string does include the reference nucleotides which are deleted. But we did notice the discrepancy with the documentation on this page: https://varnomen.hgvs.org/bg-material/simple/ where reference nucleotides are not spelled out in the format.

We did some testing at the time ... to see the behavior of the Ensembl VEP endpoints with the two different formats. We came to the conclusion that the Ensembl VEP endpoints were simply ignoring any specified nucleotides for reference allele deletions (in del or delins cases). At the time we wondered whether omitting the reference allele nucleotides would circumvent a quality check --- where a mismatch between the supplied reference allele and the reference genome which was the basis for computing the annotation might be noticed. However, no such check seemed to be occurring in the Ensembl VEP services. Random nucleotide strings (even of the wrong length) were submitted in the "fully-specified" format for hgvs where reference allele nucleotides can appear. It did not seem to cause any errors or warnings during annotation. So that testing may give us some comfort that simplifying our notation now to match the hgvs standard will be "safe".

One side effect of the work in this PR is that now additional annotation from myvariant.info will be possible ... because query strings in the fully-specified hgvs format which previously were failing when they were sent to myvariant.info (example : 17:g.7579547_7579549delGGGinsCCC) will now succeed (by using 17:g.7579547_7579549delinsCCC), so there will be additional annotation reported by genome-nexus for these categories of events. We may need to re-annotate studies to pick up this additional annotation if any myvariant.info derived fields (such as gnomAD) are stored in MAF files for our studies.

Also, we converted the previous logic into python for the gnomAD overlay tool I mentioned above. That code is here:
https://github.com/genome-nexus/annotation-tools/blob/master/variant_notation_converter.py
We will update that code in conjunction with this change. (side note: thank you for prompting us to look back at that work, because it was noticed that the python module containing the converter code had actually been left out of the PR which made the code available ... so now that has been corrected. cc: @averyniceday 

I've scanned the code changes and mock updates and they look good to me.

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java
```diff
@@ -163,20 +163,37 @@ public String genomicToHgvs(GenomicLocation genomicLocation) {
             } catch (NumberFormatException e) {
                 return null;
             }
-        } else if (var.equals("-") || var.length() == 0) {
+        } else if (var.equals("-") || var.length() == 0 || var.equals("NA") || var.contains("--")) {
```
sheridancbio, line 166: I'm not sure what a "--" substring in the variant allele means. Could someone comment on the meaning, and the appropriateness of always interpreting such a string as a deletion / indel?

sheridancbio (commented):
+1

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java
inodb: it was error in our data on cBioPortal, see also https://github.com/genome-nexus/genome-nexus/issues/276 and https://github.com/cBioPortal/datahub/issues/621. It might be fixed now tho

inodb: hmm actually i see it in our private databases when running the query, so reopened the issue

sheridancbio: I have added a note in the datahub issue that once the data has been cleaned up, the special cases for NA and '--' in the notation converter code should be removed. If those are cases which should not exist, I wonder if we can remove those cases from notationconverter now?

leexgh: I guess it doesn't hurt to have these before the issue is fixed, but we could fix it along with the nesting like Ino's comments above

PR #468, opened 2020-11-20
fix my variant info missing in BRCA2 mutations

leexgh:
**Problem**
https://www.signaldb.org/gene/BRCA2
All mutations are missing my_variant_info in annotation. 
Fix: https://github.com/genome-nexus/genome-nexus/issues/469

**Reason**
Some variants don't have `query` in `my_variant_info` response(e.g. expected to be chr7:g.140453136A>T), but we use this field to match our query(e.g. expected to be 7:g.140453136A>T)

**Solution**
Using `variant` instead

**Example**
POST:
[{"chromosome":"13","start":32972777,"end":32972777,"referenceAllele":"C","variantAllele":"G"},{"chromosome":"13","start":32972893,"end":32972893,"referenceAllele":"A","variantAllele":"-"},{"chromosome":"13","start":32890607,"end":32890607,"referenceAllele":"G","variantAllele":"T"},{"chromosome":"13","start":32911686,"end":32911686,"referenceAllele":"T","variantAllele":"-"}]
can't get my variant info, but remove the last one should be able to work

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/enricher/MyVariantInfoAnnotationEnricher.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/enricher/MyVariantInfoAnnotationEnricher.java
@@ -52,7 +52,9 @@ public class MyVariantInfoAnnotationEnricher extends BaseAnnotationEnricher
         try {
             List<MyVariantInfo> myVariantInfo = myVariantInfoService.getMyVariantInfoByAnnotation(annotations);
             Map<String, MyVariantInfo> hgvsToMyVariantInfoMap =
-                myVariantInfo.stream().collect(Collectors.toMap(MyVariantInfo::getHgvs, v -> v));
+                myVariantInfo.stream()
+                    .filter(info -> info.getHgvs() != null)
+                    .collect(Collectors.toMap(MyVariantInfo::getHgvs, v -> v));
 
             for (VariantAnnotation annotation: annotations) {
                 this.enrichAnnotationWithMyVariantInfo(annotation,
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/internal/MyVariantInfoServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/internal/MyVariantInfoServiceImpl.java
@@ -65,7 +65,9 @@ public class MyVariantInfoServiceImpl implements MyVariantInfoService
             myVariantInfos = this.getMyVariantInfoByMyVariantInfoVariant(queryVariants);
             // manually set the original hgvs variant field
             for (MyVariantInfo myVariantInfo: myVariantInfos) {
-                myVariantInfo.setHgvs(queryToVariant.get(myVariantInfo.getQuery()));
+                // myVariantInfo.getVariant() will return my_variant_info query id (e.g. chr7:g.140453136A>T)
+                // queryToVariant.get() will return genome nexus query id (e.g. 7:g.140453136A>T)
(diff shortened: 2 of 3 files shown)


onursumer:
Thanks @leexgh! I think we had a PR applying the same fix, but we closed it without merging: #457

@ao508 @sheridancbio Do you think we should still add this change (related to log) as well? https://github.com/genome-nexus/genome-nexus/pull/457/files#diff-597b7848675b8a26a2a2a88a8c0c2e5a27d3c1d8fceb1187719861c94ee93a1fR232

service/src/main/java/org/cbioportal/genome_nexus/service/enricher/MyVariantInfoAnnotationEnricher.java
```diff
@@ -52,7 +52,9 @@ public void enrich(List<VariantAnnotation> annotations) {
         try {
             List<MyVariantInfo> myVariantInfo = myVariantInfoService.getMyVariantInfoByAnnotation(annotations);
             Map<String, MyVariantInfo> hgvsToMyVariantInfoMap =
-                myVariantInfo.stream().collect(Collectors.toMap(MyVariantInfo::getHgvs, v -> v));
+                myVariantInfo.stream()
+                    .filter(info -> info.getHgvs() != null)
+                    .collect(Collectors.toMap(MyVariantInfo::getHgvs, v -> v));
```
leexgh, line 57: I've added all the changes from https://github.com/genome-nexus/genome-nexus/pull/457. Signal BRCA2 gene page (https://www.signaldb.org/gene/BRCA2) is missing all clinvar, this filter could be a quick fix for now. For variants which don't have hgvs, we don't include them in hgvsToMyVariantInfoMap

inodb (approved):
Merging this one b/c it fixes a critical error for signal

service/src/main/java/org/cbioportal/genome_nexus/service/internal/MyVariantInfoServiceImpl.java
```diff
@@ -65,7 +65,9 @@ public MyVariantInfo getMyVariantInfoByHgvsVariant(String variant)
             myVariantInfos = this.getMyVariantInfoByMyVariantInfoVariant(queryVariants);
             // manually set the original hgvs variant field
             for (MyVariantInfo myVariantInfo: myVariantInfos) {
-                myVariantInfo.setHgvs(queryToVariant.get(myVariantInfo.getQuery()));
+                // myVariantInfo.getVariant() will return my_variant_info query id (e.g. chr7:g.140453136A>T)
+                // queryToVariant.get() will return genome nexus query id (e.g. 7:g.140453136A>T)
+                myVariantInfo.setHgvs(queryToVariant.get(myVariantInfo.getVariant()));
```
inodb, line 70: as discussed it might make sense to wrap this line in try/accept such that not entire loop stops when one of these fails

service/src/main/java/org/cbioportal/genome_nexus/service/enricher/MyVariantInfoAnnotationEnricher.java
sheridancbio: Just reiterating here ... this patch is hiding the problem. *We* constructed the hgvs query string ... so failing to get it back in the response from myvariant.info is a side-effect of the problem we have parsing the (altered) myvariant.info response schema. We should remove this patch as soon as we fix the model ... otherwise we will be blind to future degradation in myvariant.info responses as their model continues to evolve.

Issue #471, opened 2020-11-25
My variant info investigation


leexgh:
**Issues noticed:**
1. Model change:

fields | type in GN | type in My Variant Info | Error message | example
-- | -- | -- | -- | --
my_variant_info.clinvar.hgvs.coding | `List<String>` | `String / List<String>` | Cannot deserialize instance of `java.util.ArrayList<java.lang.Object>` out of VALUE_STRING token at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.cbioportal.genome_nexus.model.my_variant_info_model.MyVariantInfo["clinvar"]->org.cbioportal.genome_nexus.model.my_variant_info_model.ClinVar["hgvs"]->org.cbioportal.genome_nexus.model.my_variant_info_model.Hgvs["coding"]) | **String:** 5:g.56178615A>C   **List< String >:** 6:g.152708291G>A
my_variant_info.clinvar.rcv | `List<Rcv>` | `Rcv / List<Rcv>` | Cannot deserialize instance of java.util.ArrayList<org.cbioportal.genome_nexus.model.my_variant_info_model.Rcv> out of START_OBJECT token at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.cbioportal.genome_nexus.model.my_variant_info_model.MyVariantInfo["clinvar"]->org.cbioportal.genome_nexus.model.my_variant_info_model.ClinVar["rcv"]) | **Rcv:** 7:g.55242467_55242486delinsTC  **List< Rcv >:** 6:g.152708291G>A

2. Raw data normalization
@onursumer made this fix 6 months ago.
https://github.com/genome-nexus/genome-nexus/pull/416/files#diff-76041072e6d6540d72adb88493f784395074369838915c06ac4ab495c67366e6R124-R138
This nominalization only applies to one `fetchAndCache` function(fetchAndCache(Set< String > needToFetch...), but there are two `fetchAndCache` functions.
    1. protected void fetchAndCache(Set< String > needToFetch...): https://github.com/genome-nexus/genome-nexus/blob/master/service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java#L68
    2. public T fetchAndCache(String id): https://github.com/genome-nexus/genome-nexus/blob/master/service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java#L68

[...]

PR #472, opened 2020-11-25
normalize my variant info to handle model differences

leexgh:
Fix:https://github.com/genome-nexus/genome-nexus/issues/471

- normalize raw data to allow parsing a single value as a list when reading from a web service
- revert everything from https://github.com/genome-nexus/genome-nexus/pull/468


inodb (approved):

inodb:
Thanks for fixing this 🎉 !

Issue #453, opened 2020-10-09
issue with 9:g.21994291C>T


jjgao:
I was looking at [9:g.21994291C>T](https://www.genomenexus.org/variant/9:g.21994291C%3ET) and found multiple issues:

- [x] The variant page of [9:g.21994291C>T](https://www.genomenexus.org/variant/9:g.21994291C%3ET) pick the intron mutation in _RP11-145E5.5_ instead one of the missense mutations in CDKN2A. We should work out the priority list for choosing a default transcript (maybe based on consequence and uniprot canonical isoform?)

    **Answer** The issue in this case is that for the transcript revolver we are still using Ensembl's canonical assignments instead of genome nexus' assignments. https://github.com/genome-nexus/genome-nexus/blob/master/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/CanonicalTranscriptResolver.java#L31-L35. This is why it's picking RP11 (since ensembl labels that as canonical). We do indeed pick most impactful if no canonical is found. Filed a separate issue for this to use the genome nexus canonical assignments:  https://github.com/genome-nexus/genome-nexus/issues/474

- [x] There are two transcripts for CDKN2A in the [mskcc override file](https://github.com/genome-nexus/genome-nexus-importer/blob/8d8d690a66137dff2184cedfefdac01c910bf820/data/isoform_overrides_at_mskcc.txt): `ENST00000304494` (`p16 INK4A`) and `ENST00000361570` (`p14 ARF`). Not sure how we pick a transcript in this case? (There is only one in the [uniprot override file](https://github.com/genome-nexus/genome-nexus-importer/blob/8d8d690a66137dff2184cedfefdac01c910bf820/data/isoform_overrides_uniprot.txt): `ENST00000304494` `p16 INK4A`). Note that p16 should be used as OncoKB is annotating on p16

    **Answer** Currently it picks the first one when there are two (it's picking `ENST00000304494` in this case). See [code](https://github.com/genome-nexus/genome-nexus-importer/blob/ec2789b107fa5894891546d779ab32a51adaee96/scripts/make_one_canonical_transcript_per_gene.py#L14-L15).

[...]

PR #477, opened 2020-12-04
Use genome nexus canonical assignment

### LINKED ISSUE #474: For picking a transcript use genome nexus canonical assignments, instead of ensembl
We are currently using Ensembl's `canonical` field to see what transcripts are canonical: https://github.com/genome-nexus/genome-nexus/blob/master/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/CanonicalTranscriptResolver.java#L31-L35. We should use the info about a transcript being canonical from our assignments: https://github.com/genome-nexus/genome-nexus-importer/blob/master/data/grch37_ensembl92/export/ensembl_biomart_canonical_transcripts_per_hgnc.txt

onursumer:
Fix #474


service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationSummaryServiceImpl.java
```diff
@@ -72,6 +77,10 @@ public VariantAnnotationSummaryServiceImpl(VariantAnnotationService hgvsVariantA
         this.variantClassificationResolver = variantClassificationResolver;
         this.variantTypeResolver = variantTypeResolver;
         this.exonResolver = exonResolver;
+
+        this.canonicalTranscriptResolver.setPredefinedCanonicalTranscriptIds(
+            this.ensemblService.getCanonicalTranscriptIdsBySource("genome_nexus")
```
onursumer, line 82: We should probably define enum or constant for the available isoform override sources.

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/CanonicalTranscriptResolver.java
```diff
@@ -52,4 +58,49 @@ else if (transcripts.size() > 1) {
                 variantAnnotation.getTranscriptConsequences(), variantAnnotation.getMostSevereConsequence());
         }
     }
+
+    public void setPredefinedCanonicalTranscriptIds(Set<String> predefinedCanonicalTranscriptIds) {
```
onursumer, line 62: Alternatively, we can add `persistence` module as a dependency to `component` module, and directly fetch genome nexus canonical transcript ids here.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationSummaryServiceImpl.java
```diff
@@ -72,6 +78,10 @@ public VariantAnnotationSummaryServiceImpl(VariantAnnotationService hgvsVariantA
         this.variantClassificationResolver = variantClassificationResolver;
         this.variantTypeResolver = variantTypeResolver;
         this.exonResolver = exonResolver;
+
+        this.canonicalTranscriptResolver.setPredefinedCanonicalTranscriptIds(
+            this.ensemblService.getCanonicalTranscriptIdsBySource(IsoformOverrideSource.GENOME_NEXUS)
+        );
```
inodb, line 84: Let's try to figure out what does should be. Ideally the annotation summary service should use the `isoformOverrideSource`

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/CanonicalTranscriptResolver.java
```diff
@@ -52,4 +66,45 @@ else if (transcripts.size() > 1) {
                 variantAnnotation.getTranscriptConsequences(), variantAnnotation.getMostSevereConsequence());
         }
     }
+
+    /**
+     * Relies on VEP canonical assignment.
+     */
+    public List<TranscriptConsequence> findCanonicalTranscriptCandidates(
```
inodb, line 73: @onursumer wondering if this function should maybe be private just to avoid people from accidentally using it?

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/RefSeqResolver.java
```diff
@@ -22,13 +23,35 @@ public RefSeqResolver(CanonicalTranscriptResolver canonicalTranscriptResolver)
     @Nullable
     public String resolve(VariantAnnotation variantAnnotation)
     {
-        return this.resolve(this.canonicalTranscriptResolver.resolve(variantAnnotation));
+        return this.resolve(variantAnnotation, null);
+    }
```
inodb, line 27: @onursumer should we keep these `resolve` functions without a predefined list? To me it seems like we are mostly keeping them at the moment to support our test cases. I don't think there's an actual use case for using resolve without the predefined list. Is that right?

service/src/test/java/org/cbioportal/genome_nexus/component/annotation/CanonicalTranscriptResolverTest.java
```diff
@@ -130,4 +133,39 @@ public void resolveCanonicalTranscript() throws IOException
...
+        String[] overridesWithMultipleMatch = {"ENST00000532753", "ENST00000532924", "ENST00000532880"};
+
+        // should pick the most impactful one (inframe_deletion)
+        assertEquals(
+            "ENST00000532924",
+            this.canonicalTranscriptResolver
+                .resolve(variantMockData.get("3:g.14106026_14106037del"), new HashSet<>(Arrays.asList(overridesWithMultipleMatch)))
+                .getTranscriptId()
+        );
+    }
```
inodb, line 170: Maybe good to add this example as well `9:g.21994291C>T`. Could be an integration test

inodb (commented):
Looks good! Thanks so much for fixing this!

Just some minor comments. Wondering if we can get rid of a few functions to avoid people from accidentally using them (see comments). Also would be good to add this example `9:g.21994291C>T` to the test cases (I don't think I saw it): https://github.com/genome-nexus/genome-nexus/issues/453

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/RefSeqResolver.java
onursumer: Right, currently we only use these `resolve` functions in tests. We can update the tests and remove the unused functions  if you think there is no actual use case for them.

service/src/test/java/org/cbioportal/genome_nexus/component/annotation/CanonicalTranscriptResolverTest.java
onursumer: We can add the actual data for `9:g.21994291C>T`, but the override lists here are random though.

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/RefSeqResolver.java
inodb: yeah I don't think there's a use case, so best to remove them to avoid people from accidentally using them. But could be in follow up

component/src/main/java/org/cbioportal/genome_nexus/util/IsoformOverrideSource.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public static String getOrDefault(String isoformOverrideSource) {
+        String normalizedSource = isoformOverrideSource == null ?
+            UNIPROT : isoformOverrideSource.trim().toLowerCase();
+
+        return (
+            normalizedSource.equals(ENSEMBL) ||
+            normalizedSource.equals(GENOME_NEXUS) ||
+            normalizedSource.equals(MSKCC)
+        ) ? normalizedSource : UNIPROT;
```
averyniceday, line 17: does this mean if someone passes in something "fake_source" we'll still be using UNIPROT?

averyniceday: I see, looks like that was basically existing behavior (At least in EnsemblCanonical class)

onursumer: right, currently uniprot is the default.

sheridancbio:
@onursumer I think I'm starting to understand this. I have a question, but first let me explain my current understanding ... please correct me if I have this wrong:
When annotating, we need to know what transcript should be selected from the AlterationConsequences coming out of VEP. The way to select was/is:
1. if a transcript (for an alterationConsequence) is a list of override transcripts, it gets first preference. The list of override transcripts is chosen from the isoformOverrideSource argument. If that argument is "mskcc" then we choose the mskcc override list ... otherwise we choose the "uniprot" override list.
2. if no such override is found, we compare against a list of "canonical" transcripts for the gene ... these have second preference. Currently, this canonical list is based on the transcripts marked as "canonical" in whatever version of Ensembl we are using. After this PR is merged and deployed we will be using our own (mskcc defined) list of canonical transcripts.
3. if no such canonical transcript is found in the alterationConsequence list, I am not sure what happens next. Perhaps we choose one of the alternationConsequences at random ... or by most significant effect on the protein.

My question: what is the fundamental problem we were observing? I am guessing that the problem was : over time, the set of Ensembl canonical transcripts has drifted ... and as that happened we started to diverge from the annotations which were done earlier. Does this PR attempt to "freeze" the set of canonical transcripts so that we stop seeing drift?

A general comment: I think the "canonical" word is becoming more confusing as we alter our code behavior. To me "canonical" implies that the selection is done in a way which is universally (or at least widely) acknowledged and authoritative. If I understand correctly, we are choosing here to deviate from the Ensembl canonical list ... which might be better labeled as our "chosen default" transcript when an override transcript is not available. It also makes me think that perhaps we should drop the concept of a default with possible overrides and simply represent our transcript list choices and explicit (full) lists of transcript choices. That would be a refactor outside the scope of this PR, but it does seem strange to have an mskcc (genome_nexus) list of defaults and also an mskcc list of overrides (applied on top our own defaults)

sheridancbio (approved):
We discussed the PR use cases and strategy in a group. Code changes look good to me.

sheridancbio:
We had a followup discussion about the intents and effects in this PR.
Main points of new understanding:
- The transcript override lists (mskcc and uniprot) are constructed from the mongo document (that is the purpose for accessing the Document in this PR)
- The problem before was that only some of the fields reported in the annotation_summary element were being affected by the override transcript list. The only field we know was being updated was the transcript_id field. However, other fields (such as hgvsp_short) were being resolved based on the canonical annotation being returned in the VEP response (in the alterationConsequences)
- The fix was to adjust the canonical transcript resolver so that the override list is always passed in .. and so each field resolution is done based on the final (possibly overridden) transcript
- There are now 3 levels of priority for transcript candidate selection:
1. any transcripts which are member of the override list
2. any transcripts which are marked as canonical by VEP
3. all transcripts reported by VEP (none of which are canonical)
After the candidates are identified, the single selected transcript for annotation will be from those with the most impactful protein change. If there is still ambiguity after filtering to the most impactful protein change, overall transcript length is used to select the longest transcript as the one to report.

service/src/main/java/org/cbioportal/genome_nexus/service/enricher/IsoformAnnotationEnricher.java
```diff
@@ -1,95 +1,84 @@
...
 
-            We only apply isoform overrides if the most severe consequence exists in the gene we want to apply to override to.
-            If it does not, do not touch the canonical fields for those transcripts.
-        */
-        String geneWithMostSevereConsequence = "";
-        String mostSevereConsequence = annotation.getMostSevereConsequence();
-        for (TranscriptConsequence transcript : annotation.getTranscriptConsequences()) {
-            if (transcript.getConsequenceTerms().contains(mostSevereConsequence)) {
-                geneWithMostSevereConsequence = transcript.getGeneSymbol();
-                break;
```
onursumer, line 55: @inodb we don't need this logic anymore, right?

inodb: @onursumer yeah i think we can remove it but note that the logic used to be different. I.e. to me it seems like (1) it looks for the most severe consequence and then if that is in a gene for which we have an override then we apply it. Which is different from what we are doing now: (2) filter to canonical transcripts and then get most severe event in that subset. I think for our case of annotating for the purpose of loading it into cBioPortal (2) makes more sense, b/c if it is not in a canonical transcript it means there is no associated hugo symbol and no one can even query that mutation in cBioPortal.

web/src/main/java/org/cbioportal/genome_nexus/web/IsoformOverrideController.java
```diff
@@ -1,87 +0,0 @@
-package org.cbioportal.genome_nexus.web;
```
sheridancbio, line 1: at one point we must have used this endpoint in the msk importer code. There is a property listed in importer.properties.EXAMPLE: "genomenexus.isoform_override="
However, this property is no longer in use in the importer code base.
Similarly, there is a property in genome-nexus-annotation-pipeline: annotationPipeline/src/main/resources/application-test.properties.EXAMPLE  "genomenexus.isoform_override=<HOST>/variant_annotation/isoform_override/"
but there also seems to be no references to this property in the code base any longer.
We should strip unreferenced properties in these other repositories.

web/src/main/resources/isoform_overrides_at_mskcc.txt
```diff
@@ -1,471 +0,0 @@
-enst_id	gene_name	refseq_id	ccds_id
```
sheridancbio, line 1: If this is no longer to be used, then it is correct to remove this file (and the stack which handled requests for these lists). I do wonder whether anyone in the community has need of these services though.. ?

Would it be a good idea to leave the controller in place but to simply throw an unsupportedOperation exception (or a 410 http status code) with a message such as "Our data representation for isoform overrides has been changed and this service cannot be provided without additional code refactoring. We plan to retire this service, but if you have a necessary use case for us to consider, contact the development team : (<email address>)."

web/src/main/java/org/cbioportal/genome_nexus/web/error/GlobalExceptionHandler.java
```diff
@@ -68,10 +68,4 @@
     {
         return new ResponseEntity<>(new ErrorResponse(ex.getResponseBody()), ex.getStatusCode());
     }
-
-    @ExceptionHandler(IsoformOverrideNotFoundException.class)
-    public ResponseEntity<ErrorResponse> handleIsoformOverrideNotFound(IsoformOverrideNotFoundException ex)
-    {
-        return new ResponseEntity<>(new ErrorResponse(ex.getLocalizedMessage()), HttpStatus.NOT_FOUND);
```
sheridancbio, line 75: see my previous comment. If we want to leave the controller so that a custom message goes to the community with a http status of HttpStatus.GONE we could modify this class to generate such a response (maybe with a new exception type or using UnsupportedOperation)

web/src/main/java/org/cbioportal/genome_nexus/web/IsoformOverrideController.java
onursumer: Right, we should strip properties that are no no longer in use.

Not sure if we ever used this endpoint directly though. It is possible that we've been only using the `isoformOverrideSource` option.

web/src/main/resources/isoform_overrides_at_mskcc.txt
onursumer: `IsoformOverrideController` is part of the internal API, so I am not sure if it is worth the effort. I am not even sure if this endpoint has ever been used somewhere.

service/src/main/java/org/cbioportal/genome_nexus/service/enricher/IsoformAnnotationEnricher.java
sheridancbio: I think we touched on this in our discussion yesterday. The case was two overlapping genes on opposite strands .. and also a mutation in an intergenic region near these genes. I think it would be a good idea (in the near future) to consider and formalize (in a util / service) our logic for how to select the gene we "want" to base our annotation on when there is ambiguity. Queries to the annotation service only provide chromosome coordinates, but potentially we could also accept a [optional?] gene choice which would aid in the disambiguation. But even if we do not want to do this, it would be good to have a coded (and unit tested) component which takes the ambiguous (multi-gene) transcripts coming out of VEP and applies our "pick a gene" logic to it. Once we have chosen the intended gene, we can filter the TranscriptConsequences to the set of consequences related to that gene's transcripts and then proceed to select the transcript. This way the same gene selection logic would be applied regardless of whether the canonical transcript is identified by the table in Mongo or is identified by the "canonical" element in the VEP response.
For now, when there are multiple genes in the VEP response we are changing behavior so that we prefer genes from the override set (in mongo). However, there are probably cases where this is not enough to fully disambiguate.
Also, I wonder whether every hugo symbol from the gene table has been added to the Mongo table? The final statement from @inodb above seems to imply that. If we want to rely on that, we should be careful to keep the Mongo table in sync with the gene table as there are gene table updates.

service/src/main/java/org/cbioportal/genome_nexus/service/enricher/IsoformAnnotationEnricher.java
```diff
@@ -1,95 +1,84 @@
...
-                transcript.setCanonical("1");
-            }
-            else
-            {
-                transcripts.add(transcript);
-            }
+        if (predefinedCanonicalTranscriptIds != null) {
+            transcripts = transcriptConsequences
+                .stream()
+                .filter(t -> predefinedCanonicalTranscriptIds.contains(t.getTranscriptId()))
```
sheridancbio, line 78: 👍

service/src/main/java/org/cbioportal/genome_nexus/service/internal/BaseVariantAnnotationServiceImpl.java
```diff
@@ -244,55 +245,52 @@ private EnrichmentService initPostEnrichmentService(String isoformOverrideSource
...
+            new IsoformAnnotationEnricher(isoformOverrideSource, isoformOverrideSource, ensemblService)
+        );
 
         if (fields != null && fields.contains("hotspots"))
         {
-            AnnotationEnricher enricher = new HotspotAnnotationEnricher(
-                "cancerHotspots", hotspotService, true);
-            postEnrichmentService.registerEnricher(enricher);
+            postEnrichmentService.registerEnricher(
+                new HotspotAnnotationEnricher("cancerHotspots", hotspotService, true)
```
sheridancbio, line 257: 👍

service/src/test/java/org/cbioportal/genome_nexus/service/enricher/IsoformAnnotationEnricherTest.java
```diff
@@ -0,0 +1,116 @@
...
+            "ENST00000532880",
+            canonicalTranscripts.get(1).getTranscriptId()
+        );
+
+        assertEquals(
+            "ENST00000532924",
+            canonicalTranscripts.get(2).getTranscriptId()
+        );
+    }
+}
```
sheridancbio, line 116: maybe a missing test : enrichAnnotationWithZeroOverrides ?

sheridancbio (approved):
looks good. I plan to do a little system testing as well.

Before merge, I'll wait to see if there are any thoughts / responses to my questions, such as whether it is worthwhile to leave the endpoint for overrides in place with a "we have dropped this service" message for any community users who might be using the endpoint for purposes we are not aware of.

web/src/main/resources/isoform_overrides_at_mskcc.txt
sheridancbio: Ok, that sounds good. If it was not visible as part of the public API then definitely not worth it.

web/src/main/java/org/cbioportal/genome_nexus/web/error/GlobalExceptionHandler.java
sheridancbio: resolved .. not necessary

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/CanonicalTranscriptResolver.java
```diff
@@ -52,4 +48,26 @@ else if (transcripts.size() > 1) {
...
         }
     }
+
+    /**
+     * If an isoform override have not updated the canonical transcript information,
+     * then this will rely purely on VEP canonical assignment.
+     */
+    private List<TranscriptConsequence> findCanonicalTranscriptCandidates(
+        List<TranscriptConsequence> transcriptConsequences
+    ) {
```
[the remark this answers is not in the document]
onursumer: Right, we should use something like `prettier`.

service/src/test/java/org/cbioportal/genome_nexus/service/enricher/IsoformAnnotationEnricherTest.java
onursumer: sounds good, I can add that one too.

component/src/main/java/org/cbioportal/genome_nexus/util/IsoformOverrideSource.java
```diff
@@ -0,0 +1,19 @@
...
+    public static final String GENOME_NEXUS = "genome_nexus";
+    public static final String UNIPROT = "uniprot";
+    public static final String MSKCC = "mskcc";
+
+    public static String getOrDefault(String isoformOverrideSource) {
+        String normalizedSource = isoformOverrideSource == null ?
+            UNIPROT : isoformOverrideSource.trim().toLowerCase();
+
+        return (
+            normalizedSource.equals(ENSEMBL) ||
```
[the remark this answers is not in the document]
onursumer: right, but not sure if we really need anything other than uniprot and mskcc.

sheridancbio:
Additional testing performed, and expected behavior observed. Summary:
all annotation endpoint variants visited with both uniprot and mskcc isoform overrides for the following queries:
**class1**
queries which vary by selected transcript .. in a non-varying and also a varying position:
GRCh37  6       157099136       157099136  C>G    --> p.L25V (ENST00000350026.5 ARID1B-009 ) UNIPROT
GRCh37  6       157099136       157099136  C>G    --> p.L25V (ENST00000346085.5 ARID1B-001) MSKCC
3,52598099,52598099,G,A --> p.A1281V ( ENST00000296302.7 ) UNIPROT
3,52598099,52598099,G,A --> p.A1256V ( ENST00000394830.3 ) MSKCC
(changes in order and count of variants also explored)
**class2**
the problematic case related to CDKN2B / CDKN2B-AS1 / RP11-145E5.5
9:g.21994291C>T
**class3**
a case where VEP annotation yields no transcripts marked as canonical, but there is still a choice to make:
17:g.5676040C>A

Issue #487, opened 2021-02-07
Variant Classification for Indels is Inconsistent


as1000:
Currently, the "variantClassification" for in frame indels is either "In_Frame_Del" or "In_Frame_Ins":
https://www.genomenexus.org/annotation/7:g.55242468_55242481delinsAC?fields=hotspots%2Cmy_variant_info%2Cptms%2Cannotation_summary

Likewise, that for frameshift indels is either "Frame_Shift_Del" or "Frame_Shift_Ins"
https://www.genomenexus.org/annotation/17:g.7579854_7579858delinsT?fields=hotspots%2Cmy_variant_info%2Cptms%2Cannotation_summary

The arises due to the way VariantType is resolved ("resolveVariantType"):
https://github.com/genome-nexus/genome-nexus/blob/184aa763e2d386274a29455b06ae6582966c157c/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/VariantTypeResolver.java

The resulting variantType is used to determine the variantClassification in VariantClassificationResolver:
https://github.com/genome-nexus/genome-nexus/blob/184aa763e2d386274a29455b06ae6582966c157c/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/VariantClassificationResolver.java

To fix this issue, a "DELINS" classification should be added to resolveVariantType.

PR #488, opened 2021-02-08
Resolve_VariantClassification_for_Indels

as1000:
Addressing issue #487

### DIFF
--- a/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/VariantClassificationResolver.java
+++ b/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/VariantClassificationResolver.java
@@ -117,6 +117,9 @@ public class VariantClassificationResolver
             else if (variantType.equals("INS")) {
                 return "Frame_Shift_Ins";
             }
+            else if (variantType.equals("DELINS")) {
+                return "Frame_Shift_Delins";
+            }
         }
         else if ((variant.equals("protein_altering_variant") || variant.equals("coding_sequence_variant")) &&
             isInframe)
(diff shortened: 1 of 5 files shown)


as1000:
The issue with the Travis CI build is that 13:g.28608258_28608275del has the wrong "allele_string". It should be CATATTCATATTCTCTGA/- but instead, it is CATATTCATATTCTCTGA/GGGGTGGGGGGG for some reason. This makes the updated VariantTypeResolver classify this mutation as an "DELINS", as it rightfully should.

https://github.com/genome-nexus/genome-nexus/blob/master/service/src/test/java/org/cbioportal/genome_nexus/component/annotation/VariantTypeResolverTest.java 

`        // allele_string: "CATATTCATATTCTCTGA/GGGGTGGGGGGG" => expected: DEL
        assertEquals(
            "Variant type for 13:g.28608258_28608275del should be DEL",
            "DEL",
            this.variantTypeResolver.resolve(variantMockData.get("13:g.28608258_28608275del"))
        );`

Other weird things are occurring with 13:g.28608258_28608275del, as well. It's Hgvs Short is being classified as a DELINS (I'm guessing this is a consequence of the improper "allele_string").

https://www.genomenexus.org/variant/13:g.28608258_28608275del

However, the json summary for this variant classifies the "allele_string" correctly as CATATTCATATTCTCTGA/-.

https://www.genomenexus.org/annotation/13:g.28608258_28608275del?fields=hotspots%2Cmy_variant_info%2Cptms%2Cannotation_summary

inodb (commented):
Thanks so much @as1000 !

Looking at this some more. Although it makes a lot of sense I'm not sure if this follows the MAF spec:

https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/

It still makes sense to have some field in the JSON to indicate delins as you propose but im a bit hesitant to add new vocabulary to official MAF fields like Variant_Classification and Variant_Type. Not sure e.g. how GDC shows those mutations in their MAFs. Could you check how they handle these cases https://portal.gdc.cancer.gov/? Thanks!

PR #492, opened 2021-02-17
Create mongodb index by region to speed up signal queries

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/8318

Build signal query index by `chromosome`, `start` and `end`. 
10k queries take more than 10 minutes before, and take about 30s now.

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/enricher/SignalAnnotationEnricher.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/enricher/SignalAnnotationEnricher.java
@@ -1,9 +1,10 @@
 package org.cbioportal.genome_nexus.service.enricher;
 
+import org.cbioportal.genome_nexus.component.annotation.GenomicLocationResolver;
+import org.cbioportal.genome_nexus.model.GenomicLocation;
 import org.cbioportal.genome_nexus.model.SignalAnnotation;
 import org.cbioportal.genome_nexus.model.SignalMutation;
 import org.cbioportal.genome_nexus.model.VariantAnnotation;
-import org.cbioportal.genome_nexus.service.AnnotationEnricher;
 import org.cbioportal.genome_nexus.service.SignalMutationService;
 
 import java.util.List;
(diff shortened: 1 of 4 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/enricher/SignalAnnotationEnricher.java
```diff
@@ -23,13 +24,9 @@ public SignalAnnotationEnricher(
...
     public void enrich(VariantAnnotation annotation) {
         SignalAnnotation signalAnnotation = new SignalAnnotation();
-        String hgvsg = annotation.getHgvsg();
-
-        if (hgvsg != null) {
-            List<SignalMutation> mutations = this.signalMutationService.getSignalMutationsByHgvsg(hgvsg);
-            signalAnnotation.setAnnotation(mutations);
-        }
-
+        GenomicLocation genomicLocation = GenomicLocationResolver.resolve(annotation);
```
onursumer, line 27: Let's revert the changes in `GenomicLocationResolver` and create an instance of `GenomicLocationResolver` in the constructor instead. Something like this should work:

```Java
private final GenomicLocationResolver genomicLocationResolver;

public SignalAnnotationEnricher(...) {
    ...
    this.genomicLocationResolver = new GenomicLocationResolver()
}

```

Then you can simply do this here:
```suggestion
        GenomicLocation genomicLocation = this.genomicLocationResolver.resolve(annotation);
```

onursumer (changes requested):

service/src/main/java/org/cbioportal/genome_nexus/service/enricher/SignalAnnotationEnricher.java
leexgh: @onursumer Sure, updated. Why static is not good here?

onursumer: Thanks! Static methods may break component autowiring. Also, I think only dynamic methods can be mocked in test classes.

onursumer (approved):
LGTM

PR #496, opened 2021-03-23
Update signal model to include germline biallelic

leexgh:
Part of: https://github.com/knowledgesystems/signal/issues/112

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/StatsByTumorType.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/StatsByTumorType.java
@@ -26,21 +26,24 @@ public class StatsByTumorType {
     private Double msiScore;
 
     @Field("n_with_sig")
-    private Integer nWithSig;
+    private Integer numberWithSig;
 
     @Field("hrd_score")
     private HrdScore hrdScore;
+
+    @Field("n_germline_homozygous")
+    private Integer numberOfGermlineHomozygous;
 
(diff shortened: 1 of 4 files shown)


web/src/main/java/org/cbioportal/genome_nexus/web/mixin/StatsByTumorTypeMixin.java
```diff
@@ -26,8 +26,11 @@
     private Double msiScore;
 
     @ApiModelProperty("Number With Signature")
```
onursumer, line 28: This should be ideally `Number of Variants with Signature` or something like that, right? But I guess we can think about it later.

onursumer (approved):

model/src/main/java/org/cbioportal/genome_nexus/model/StatsByTumorType.java
```diff
@@ -26,21 +26,24 @@
     private Double msiScore;
 
     @Field("n_with_sig")
-    private Integer nWithSig;
+    private Integer numberWithSig;
```
leexgh, line 29: The main reason I change this is somehow `nWithSig` will be changed to `nwithSig` in model, as well as `nGermlineHomozygous` (changed to `ngermlineHomozygous`), but other fields like `nCancerTypeCount` or `fBiallelic` are fine. I tried adding JsonProperty but didn't help, so I have to rename them.

PR #498, opened 2021-03-31
Validate Reference_Allele string when annotating (if present)

sheridancbio:
## Problem
When annotation api requests specify deleted nucleotides, genomenexus currently does not check whether those nucleotides match the content of the reference genome assembly which is in use by VEP to make variant effect predictions. This is because of simplification of the query format before passing the request to VEP.

When a genome nexus server uses the public Ensembl vep annotation endpoints for predictions, queries are submitted in the (standard simplified) HGVS format. This format no longer includes the nucleotides which are deleted, for example with this deletion-insertion variants: /annotation/genomic/5,138163255,138163256,CC,TT

the query which is made Ensembl does not mention the replaced "CC" nucleotides but is encoded like this:
5:g.138163255_138163256delinsTT
So a predicted effect is generated using the genome assembly available to VEP (which has sequence "TC" rather than "CC" at these positions.

When a genome nexus server uses a local deployment of the [VEP command line tool](https://github.com/genome-nexus/genome-nexus-vep) the query for the same case is formatted in a REST-style region format which also does not mention the deleted nucleotides:
5:138163255-138163256 :-/TT

There are subtle differences in output response between these two configured VEP service providers when invariant nucleotides are part of the "replacement". In other words, the results can differ if the query "replaces" a T with a T in the first position of a 2 NT delins where the second position is a genuine replacement (C>T).

To avoid these inconsistencies, we have decided to require that any query which specifies a reference allele must be validated to make sure the provided reference allele matches the genome assembly in use by VEP. During annotation of our own studies and variants we will provide the reference allele, increasing reliability and standardization in results.

## Strategy
Two services are available for providing annotation. One using HGVS format making requests to the Ensembl /annotation endpoint, One using REST-style region format to make requests to a locally deployed VEP web service. Calls to these services will be redirected to a higher level service which both makes the appropriate service call for annotation, and then verifies that the annotation returned from the service contains a reference genome sequence which matches the input Reference_Allele sequence. For some types of variants (SNV) this information will always be available and validation can be done with one request. For other types of variants (delins) the annotation returned from the service may not contain the full reference genome sequence if the extent of the variation has been reduced (as discussed above when part of the "replacement" is actually invariant). In those cases a subsequent request to the annotation service provider will be made for a full deletion of the coordinates covered by the Reference_Allele. Then the full reference genome sequence will be compared to the Reference_Allele. When it is not a match, the service will report a failure to annotate.

### DIFF
--- a/component/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
+++ b/component/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
@@ -138,6 +138,41 @@ public class GenomicVariantUtil {
         return Pattern.matches("^[\\w\\-\\./]+(\\.maf)|^[\\w\\-\\./]+(\\.tsv)|^[\\w\\-\\./]+(\\.txt)$$", file);
     }
 
+    // Hgvs formatted variants of several types are parsed for provided Reference Alleles
+    // empty string is returned if no reference allele can be extracted
+    public static String providedReferenceAlleleFromHgvs(String hgvs) {
+        // Substitutions require a single nucleotide reference sequence:
+        // 1g.123456A>G  (chromosome 1, genomic position 123456, reference sequence A, substituted with G)
+        // ref: http://varnomen.hgvs.org/recommendations/DNA/variant/substitution/
+        String match = getPattern("(?<=[^ATGC])[ATGC]>", hgvs);
+        if (match != null && match.trim().length() > 0) {
+            return match.substring(0, 1);
+        }
+        // Deletion-Insertions are recommended to *not* include specified deleted seqeunce/Reference Allele, but it is not prohibited.
+        // Examples showing specified deleted sequence/ReferenceAllele are mentioned here : http://varnomen.hgvs.org/recommendations/DNA/variant/delins/
+        // 1g.123456_123457delAAinsTT (chromosome 1, genomic positions 123456-123457, reference sequence AA deleted, sequence TT inserted in place)
+        // recommended representation omits the deleted nucleotides "1g.123456_123457delinsTT"
+        match = getPattern("del[ATGC]*ins", hgvs);
+        if (match != null && match.trim().length() > 0) {
+            return match.trim().substring(3, match.trim().length() - 3);
+        }
+        // Deletions are recommended to *not* include specified deleted seqeunce/Reference Allele, but it is not prohibited.
+        // An example showing a specified deletion is mentioned here : http://varnomen.hgvs.org/recommendations/DNA/variant/alleles/
+        // 1g.123456_123457delAA (chromosome 1, genomic positions 123456-123457, reference sequence AA deleted)
+        // recommended representation omits the deleted nucleotides "1g.123456_123457del"
+        match = getPattern("del[ATGC]*", hgvs);
+        if (match != null && match.trim().length() > 0) {
+            return match.trim().substring(3);
+        }
+        // Duplications are not supported by this system, but would be handled with logic similar to deletions
+        // Examples showing specified deleted sequence/ReferenceAllele are mentioned here : http://varnomen.hgvs.org/recommendations/DNA/variant/duplication/
+        // 1g.123456_123457dupAA (chromosome 1, genomic positions 123456-123457, reference sequence AA duplicated)
+        // Hgvs repeats and inversions are not supported by this system and this method does not attempt to parse these formats
+        //
+        // no reference allele was found
+        return "";
+    }
+
     // postcondition: returns a substring of hgvs that matched to the regex, or null if not matched
     private static String getPattern(String regex, String hgvs) {
         Pattern p = Pattern.compile(regex);
--- a/component/src/test/java/org/cbioportal/genome_nexus/util/GenomicVariantUtilTest.java
+++ b/component/src/test/java/org/cbioportal/genome_nexus/util/GenomicVariantUtilTest.java
@@ -2,7 +2,7 @@ package org.cbioportal.genome_nexus.util;
 
 import static org.junit.Assert.assertEquals;
 
-import java.util.ArrayList;
+import java.util.*;
 
 import org.junit.Test;
 
@@ -186,4 +186,46 @@ public class GenomicVariantUtilTest {
         assertEquals("G", two.getRef());
         assertEquals("T", two.getAlt());
     }
+
+    @Test
+    public void testProvidedReferenceAlleleFromHgvs() {
+        // The tested function only extracts specified reference alleles. Position ranges are not examined for correct count.
+        String NONE = "";
+        LinkedHashMap<String, String> testCases = new LinkedHashMap<String, String>();
+        // Substitution cases
+        testCases.put("1g.1A>G", "A");
+        testCases.put("1g.1T>G", "T");
+        testCases.put("1g.1C>G", "C");
+        testCases.put("1g.1G>G", "G");
+        testCases.put("1g.1G>G ", "G"); // trailing whitespace is acceptable
+        testCases.put("1g.1C>", "C"); // reference allele was specified, even if required tumor seq allele was missing
+        testCases.put("1g.1AA>G", NONE); // too many nucleotides in reference allele
+        testCases.put("1g.1>G", NONE); // reference allele missing
+        testCases.put("1g.1>", NONE); // both alleles missing
+        // Deletion-Insertion cases
+        testCases.put("1g.1delinsT", NONE);
+        testCases.put("1g.1delinsT ", NONE);
+        testCases.put("1g.1delAinsT", "A");
+        testCases.put("1g.1delAinsT ", "A");
+        testCases.put("1g.1delAins", "A"); // tumor seq allele
+        testCases.put("1g.1insTdelA", "A"); // although order of ins-del is invalid, the specified reference allele is returned
+        testCases.put("1g.1_2delAAinsTT", "AA");
+        testCases.put("1g.1_3delAAAinsT", "AAA");
+        testCases.put("1g.1_2delinsTT", NONE);
+        testCases.put("1g.1_2delAinsTT", "A"); // although nucleotide count does not match reference allele size, the tested function will extract the specified Reference Allele
+        testCases.put("1g.1delAAinsTT", "AA"); // although nucleotide count does not match reference allele size, the tested function will extract the specified Reference Allele
+        // Deletion cases
+        testCases.put("1g.1del", NONE);
+        testCases.put("1g.1del ", NONE);
+        testCases.put("1g.1delA", "A");
+        testCases.put("1g.1delA ", "A");
+        testCases.put("1g.1_2delAA", "AA");
+        testCases.put("1g.1_3delAAA", "AAA");
+        testCases.put("1g.1_2del", NONE);
+        testCases.put("1g.1_2delA", "A");
+        testCases.put("1g.1delAA", "AA");
+        // if supported, duplication cases would have the same rules / test cases as deletion
+        testCases.forEach((hgvs, expectedValue) -> assertEquals("for test case " + hgvs, expectedValue, GenomicVariantUtil.providedReferenceAlleleFromHgvs(hgvs)));
+    }
+
 }
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/GenomicLocation.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/GenomicLocation.java
@@ -63,4 +63,66 @@ public class GenomicLocation
     public String toString() {
         return this.getChromosome() + "," + this.getStart() + "," + this.getEnd() + "," + this.getReferenceAllele() + "," + this.getVariantAllele();
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (!(obj instanceof GenomicLocation)) {
+            return false;
+        }
+        GenomicLocation gl = (GenomicLocation)obj;
+        // compare start
+        if (this.start == null && gl.start != null) {
+            return false;
+        }
+        if (!this.start.equals(gl.start)) {
+            return false;
+        }
+        // compare end
+        if (this.end == null && gl.end != null) {
+            return false;
+        }
+        if (!this.end.equals(gl.end)) {
+            return false;
+        }
+        // compare chromosome
+        if (this.chromosome == null && gl.chromosome != null) {
+            return false;
+        }
+        if (!this.chromosome.equals(gl.chromosome)) {
+            return false;
+        }
+        // compare referenceAllele
+        if (this.referenceAllele == null && gl.referenceAllele != null) {
+            return false;
+        }
+        if (!this.referenceAllele.equals(gl.referenceAllele)) {
+            return false;
+        }
+        // compare variantAllele
+        if (this.variantAllele == null && gl.variantAllele != null) {
+            return false;
+        }
+        if (!this.variantAllele.equals(gl.variantAllele)) {
+            return false;
+        }
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        if (this == null) {
+            return 0;
+        }
+        int hash = 7;
+        hash =  31 * hash + (start == null ? 0 : start.hashCode());
+        hash =  31 * hash + (end == null ? 0 : end.hashCode());
+        hash =  31 * hash + (chromosome == null ? 0 : chromosome.hashCode());
+        hash =  31 * hash + (referenceAllele == null ? 0 : referenceAllele.hashCode());
+        hash =  31 * hash + (variantAllele == null ? 0 : variantAllele.hashCode());
+        return hash;
+    }
+
 }
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/internal/GenomicLocationAnnotationServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/internal/GenomicLocationAnnotationServiceImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015 Memorial Sloan-Kettering Cancer Center.
+ * Copyright (c) 2015 - 2021 Memorial Sloan-Kettering Cancer Center.
  *
  * This library is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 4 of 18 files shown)


sheridancbio:
integration tests are currently being debugged. Note that resetting to a PR with only minimal changes from the current head of master led to all continuous integration tests passing

service/src/main/java/org/cbioportal/genome_nexus/service/internal/GenomicLocationAnnotationServiceImpl.java
```diff
@@ -88,20 +88,36 @@ public GenomicLocationAnnotationServiceImpl(CachedVariantRegionAnnotationFetcher
...
     }
 
     @Override
     public List<VariantAnnotation> getAnnotations(List<GenomicLocation> genomicLocations)
     {
-        return this.variantAnnotationService.getAnnotations(this.notationConverter.genomicToEnsemblRestRegion(genomicLocations));
+        Map<String, String> convertedVarsToOrigVarQueryMap = mapConvertedVarsToOrigVarQuery(genomicLocations);
+        List<VariantAnnotation> variantAnnotations = new ArrayList<>();
+        this.variantAnnotationService.getAnnotations(
+                /////////////    WHY IS THE NEXT LINE DIFFERENT LOWER DOWN IN THE CODE? ( this.genomicLocationsToVariantFormats.convert(genomicLocations), )
```
ao508, line 112: was this added here?

which line lower in the code is it referring to? 

maybe a question that @onursumer @leexgh might be able to answer?

service/src/main/java/org/cbioportal/genome_nexus/service/internal/SelectedAnnotationServiceImpl.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @Override
+    public VariantAnnotation getAnnotation(
+            String variant,
+            String isoformOverrideSource,
+            Map<String, String> token,
+            List<String> fields) throws VariantAnnotationNotFoundException, VariantAnnotationWebServiceException {
+        if (needToConvertHgvsToRegionForAnnotation(variant)) {
+            GenomicLocation variantAsGenomicLocation = notationConverter.hgvsgToGenomicLocation(variant);  
+            // TODO: we should provide a getAnnotation() call to genomicLocationAnnotationService which accepts GenomicLocation
```
ao508, line 82: perhaps the other Genome Nexus developers can help answer these two questions?

service/src/main/java/org/cbioportal/genome_nexus/service/internal/SelectedAnnotationServiceImpl.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        if (!isHgvsFormat(variant)) {
+            return false;
+        }
+        return true;
+    }
+
+    private boolean needToConvertHgvsToRegionForAnnotation(List<String> variants) throws VariantAnnotationQueryMixedFormatException {
+        if (!isRegionAnnotationEnabled) {
+            return false; // without a local gn_vep service processing region format queries, all variants are processeed via hgvsAnnotationService
```
ao508, line 122: typo in comment: "processeed"

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedGenomicLocationAnnotationServiceImpl.java
```diff
@@ -0,0 +1,200 @@
...
+        List<VariantAnnotation> annotations = genomicLocationAnnotationService.getAnnotations(genomicLocations, isoformOverrideSource, token, fields);
+        for (int index = 0; index < annotations.size(); index = index + 1) {
+            VariantAnnotation annotation = annotations.get(index);
+            VariantAnnotation verifiedAnnotation = verifyOrFailAnnotation(annotation);
+            annotations.set(index, verifiedAnnotation);
+        }
+        return annotations;
+    }
+
+    private VariantAnnotation verifyOrFailAnnotation(VariantAnnotation annotation)
```
ao508, line 124: quick question here: would this handle the case where the original input query's reference allele has a common prefix with the tumor allele?

I guess what I'm wondering is whether an adjusted reference allele from the input query would still match the reference allele resolved from the genome? if the coordinates were adjusted also then I think this should probably be taken care of already but just wanted to double check

ao508: actually, is this even applicable in this context since "Verified" annotation service implementations appear to require a local running vep service? or am i mistaken?

service/src/test/java/org/cbioportal/genome_nexus/service/internal/VerifiedGenomicLocationAnnotationServiceTest.java
```diff
@@ -0,0 +1,462 @@
...
+
+    private void setUpQueryToStubMaps(String variantQuery, boolean successfullyAnnotated, String alleleString)
+    {
+        variantToVepSuccessfullyAnnotated.put(variantQuery, successfullyAnnotated);
+        variantToVepAlleleString.put(variantQuery, alleleString);
+    }
+
+    private void setUpQueryToStubMaps()
+    {
+        // VEP responses for these test cases were extraced from queries to http://genie.genomenexus.org/annotation/genomic/<variant>
```
ao508, line 151: typo in comment: "extraced"

ao508 (commented):
Left some comments but for the most part I think the code looks good. 

One thing in particular that I'm concerned about is the override `equals(0` method for the GenomicLocation object. I've left a comment there with more details.

component/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -138,6 +138,42 @@ public static boolean isMafFile(String file) {
         return Pattern.matches("^[\\w\\-\\./]+(\\.maf)|^[\\w\\-\\./]+(\\.tsv)|^[\\w\\-\\./]+(\\.txt)$$", file);
     }
 
+    // Hgvs formatted variants of several types are parsed for provided Reference Alleles
+    // empty string is returned if no reference allele can be extracted
+    public static String providedReferenceAlleleFromHgvs(String hgvs) {
+        // Substitutions require a single nucleotide reference sequence:
+        // 1g.123456A>G  (chromosome 1, genomic position 123456, reference sequence A, substituted with G)
+        // ref: http://varnomen.hgvs.org/recommendations/DNA/variant/substitution/
+        String match = getPattern("(?<=[^ATGC])[ATGC]>", hgvs);
```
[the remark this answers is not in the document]
sheridancbio: I'm not sure how often such requests are received by the endpoints at genomenexus.org. The import pipeline uses the genomic endpoints, and the mutations tab of the query response page in cbioportal also makes queries using genomicLocation objects posted to "https://v1.genomenexus.org/annotation/genomic". So it may be that normal operations with portal imports and websites do not use the /annotation endpoint at all. However, the genome nexus server code was recently changed so that when sending requests to Ensembl's VEP webapi using HGVS, we standardize our requests to a format which omits the reference allele string --- and this seems to be the recommended standard now here : https://varnomen.hgvs.org/ so maybe we should expect/support the (simplified) format.

component/src/test/java/org/cbioportal/genome_nexus/util/GenomicVariantUtilTest.java
```diff
@@ -186,4 +186,46 @@ public void testMafToGenomicVariant() {
...
+        testCases.put("1g.1delinsT ", NONE);
+        testCases.put("1g.1delAinsT", "A");
+        testCases.put("1g.1delAinsT ", "A");
+        testCases.put("1g.1delAins", "A"); // tumor seq allele
+        testCases.put("1g.1insTdelA", "A"); // although order of ins-del is invalid, the specified reference allele is returned
+        testCases.put("1g.1_2delAAinsTT", "AA");
+        testCases.put("1g.1_3delAAAinsT", "AAA");
+        testCases.put("1g.1_2delinsTT", NONE);
+        testCases.put("1g.1_2delAinsTT", "A"); // although nucleotide count does not match reference allele size, the tested function will extract the specified Reference Allele
+        testCases.put("1g.1delAAinsTT", "AA"); // although nucleotide count does not match reference allele size, the tested function will extract the specified Reference Allele
```
inodb, line 216: these are nice but as mentioned above the reference allele in new formats of hgvs is removed for delins. Nice to have tho

service/src/test/java/org/cbioportal/genome_nexus/service/internal/VerifiedGenomicLocationAnnotationServiceTest.java
```diff
@@ -0,0 +1,462 @@
...
+            glInsertionDeletions.add(new VariantTestCase("5,138163255,138163256,TC,TT", true, "C/T", "2nt deletion with RefAllele, 2nt insertion, partial change"));
+            glInsertionDeletions.add(new VariantTestCase("5,138163255,138163256,TC,GG", true, "TC/GG", "2nt deletion with RefAllele, 2nt insertion, full change"));
+            glInsertionDeletions.add(new VariantTestCase("5,138163255,138163256,CC,TC", false, null, "2nt deletion with discrepant RefAllele, 2nt insertion no change")); // note : different result than ensembl
+            glInsertionDeletions.add(new VariantTestCase("5,138163255,138163256,CC,TT", false, null, "2nt deletion with discrepant RefAllele, 2nt insertion, partial change"));
+            glInsertionDeletions.add(new VariantTestCase("5,138163255,138163256,CC,GG", false, null, "2nt deletion with discrepant RefAllele, 2nt insertion, full change"));
+            glInsertionDeletions.add(new VariantTestCase("5,138163255,138163256,CCCC,TT", false, null, "2nt deletion with invalid RefAllele, 2nt insertion"));
```
inodb, line 139: nice amount of corner cases covered 🥇

component/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -138,6 +138,42 @@ public static boolean isMafFile(String file) {
...
+        if (match != null && match.trim().length() > 0) {
+            return match.substring(0, 1);
+        }
+        // Deletion-Insertions are recommended to *not* include specified deleted seqeunce/Reference Allele, but it is not prohibited.
+        // Examples showing specified deleted sequence/ReferenceAllele are mentioned here : http://varnomen.hgvs.org/recommendations/DNA/variant/delins/
+        // 1g.123456_123457delAAinsTT (chromosome 1, genomic positions 123456-123457, reference sequence AA deleted, sequence TT inserted in place)
+        // recommended representation omits the deleted nucleotides "1g.123456_123457delinsTT"
+        match = getPattern("del[ATGC]*ins", hgvs);
+        if (match != null && match.trim().length() > 5) {
```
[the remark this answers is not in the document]
sheridancbio: Yes, this is a little hacky. Here is the explanation:
This function makes three attempts to parse a ReferenceAllele string from HGVS. Each case "finds" the piece containing the reference allele with little flanking characters that need to be stripped. The idea behind the minimum length requirements was to avoid any chance of a IndexOutOfBoundsException getting thrown from String.substring(). In this middle case we are stripping 3 characters off of the start of the matched string and 3 characters off of the end of the matched string ... so we will be "safe" if we never attempt to strip from a string which is shorter than 6 characters. However, due to the nature of the regular expression, it is not possible for the matched string to be less than 6 characters so this check is not really needed here. Instead, we can be "safe" just by making sure the string "match" is not empty (length() > 0). I'll change to that here ... and the same goes for the stripping of 3 characters from the deletion case (see next comment)

component/src/main/java/org/cbioportal/genome_nexus/util/GenomicVariantUtil.java
```diff
@@ -138,6 +138,42 @@ public static boolean isMafFile(String file) {
...
+            int stringLength = match.trim().length();
+            return match.trim().substring(3, match.trim().length() - 3);
+        }
+        // Deletions are recommended to *not* include specified deleted seqeunce/Reference Allele, but it is not prohibited.
+        // An example showing a specified deletion is mentioned here : http://varnomen.hgvs.org/recommendations/DNA/variant/alleles/
+        // 1g.123456_123457delAA (chromosome 1, genomic positions 123456-123457, reference sequence AA deleted)
+        // recommended representation omits the deleted nucleotides "1g.123456_123457del"
+        match = getPattern("del[ATGC]*", hgvs);
+        if (match != null && match.trim().length() > 2) {
```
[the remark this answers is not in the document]
sheridancbio: see previous comment. This stripped three characters from the matched string, so the check was to make sure we didn't try to do that from a string of less than length 3. But it is not necessary to check.

model/src/main/java/org/cbioportal/genome_nexus/model/GenomicLocation.java
```diff
@@ -63,4 +63,35 @@ public void setOriginalInput(String originalInput) {
...
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (!(obj instanceof GenomicLocation)) {
+            return false;
+        }
+        GenomicLocation gl = (GenomicLocation)obj;
+        return this.start.equals(gl.start) &&
```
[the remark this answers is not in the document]
sheridancbio: good catch. I will add that

service/src/main/java/org/cbioportal/genome_nexus/service/internal/GenomicLocationAnnotationServiceImpl.java
sheridancbio: This comment was left here accidentally. I will remove it, and ask the question with context in a more understandable way.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/SelectedAnnotationServiceImpl.java
sheridancbio: that would be great. A little more detail for line 82 : the genomicLocationAnnotationService accepts *string* representations of GenomicLocation objects, but maybe it would be more direct / unified if we did not need to convert to a string here before sending the request. (which will get converted back into GenomicLocation in the lower level service)
But the bigger question is whether we wish to support non-HGVS queries coming in to the /annotation endpoint. It seems like we used to, and kind of still do .. but it is not clear whether it still works. It seems to be that when the backend is genome-nexus-vep we would need to convert these other supported formats into region format if we want to submit them to the VEP command line tool.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/SelectedAnnotationServiceImpl.java
sheridancbio: 👍

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedGenomicLocationAnnotationServiceImpl.java
sheridancbio: First clarification : the Verified annotation services always run, regardless of whether genome-nexus is using a local deployment of vep (genome-nexus-vep) or is using the public ensembl VEP API for remote prediction generation. When using public ensembl predictions, all queries are converted into HGVS format and are submitted to the /vep/human/hgvs/{VARIANT_GOES_HERE} endpoint at ensembl (or wherever the deployer has set the property `vep.url` to point to. This seems to always use the GET method endpoint for fetching from Ensembl (VEPDataFetcher). When using a local VEP deployment for predictions, all queries are converted into REST-like Region format for processing by the VEP command line tool in the genome-nexus-vep image. Note that in both cases, all reference allele information is lost : Region format does not include reference allele strings, and the NotationConverter logic was altered in #466 to strip reference allele content when converting to GenomicLocation to hgvs (for all types except SNP/substitutions which *do* include the replaced nucleotide), and I believe other service layer logic was added when processing hgvs queries so that reference allele specifications are removed before forwarding queries to Ensembl. So this PR is handling the verification of any user specified reference allele string (regardless of backend) through a check of the allele_string member of the returned VariantAnnotation object from whichever backend is configured to make predictions.

Responding to the question of what happens when there is a common prefix:
Users can submit 2 formats : GenomicLocation and HGVS. GenomicLocation always includes both reference allele and variant allele specifications (sometimes '-' for empty). We are accepting HGVS query strings which might or might not contain the reference allele. So in theory we should always be able to pre-process GenomicLocation queries which contain a common prefix (like "AT" -> "AC") and rewrite them as normalized queries which adjust the start position accordingly and issue a query like "T" -> "C" to the VEP backend (whichever one is in use). We will report the predicted variant allele for the normalized query string, but we also include the original_variant_query field to show the raw input. HGVS queries cannot always be pre-processed, because the user may not have provided reference allele string specifiers (for del or delins types). We could (I think we do, but I will check) preprocess HGVS requests to remove common prefixes when the reference allele is provided. When it is not provided, we submit the full query as given to the backend VEP. If a delins "replaces" a prefix substring with an identical replacement variant substring prefix (for example : user query string is 5:g.138163255_138163256delinsTT but the reference assembly sequence in the VEP database for these two positions has sequence "TC" --- having a common prefix of "T")  and then we will receive a response from the backend. This is the case that motivated this work in the first place - the problem is that the local VEP backend will respond with something like "TC->TT" but the public ensembl backend will respond with something like "C->T". This is why this issue was added : #515 --- because in order for these cases to receive a normalized response we would need to post-process or resubmit responses which contain a shared prefix.

In summary : this PR does not modify the preexisting logic which strips common prefix strings from alleles (GenomicLocations have common prefixes removed in NotationConverter.normalizeGenomicLocation(), I am not certain where the HGVS prefix removal code is), so any prefix removal logic should still be applied as before. This PR adds a layer of "reviewing the results" above the AnnotationService classes which were being called before to simply check whether the user's specified reference allele string matches what comes back in the response's "allele_string" data member. (possibly submitting a follow up request if necessary). Both the Ensembl / HGVS annotation server and the genome-nexus-vep / GenomicLocationAnnotationService are covered by the check.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/SelectedAnnotationServiceImpl.java
sheridancbio: fixed

Issue #443, opened 2020-09-08
Mapping from transcript id to uniprot accession


leexgh:
Part of: #440 
Current UniProt endpoints use UniProt accession as key, we want to query PTMs by transcript id as well.

Approach 1: 
- try mapping with UniProt mapping tool with hugo symbols https://www.uniprot.org/uploadlists/
- compare sequences

Approach 2:
- mapping by ccds id. The ccds website has a mapping file and we could use it to map our data

Approach 3:
- mapping by protein sequence: ENSP -> Ensembl sequence = UniProt sequence -> UniProt id
  - Sequence from Ensembl:
    - From api: https://grch37.rest.ensembl.org/sequence/id/ENSP00000288602?content-type=application/json
    - GRCh37: ftp://ftp.ensembl.org/pub/grch37/release-101/fasta/homo_sapiens/pep/
    - GRCh38: ftp://ftp.ensembl.org/pub/release-101/fasta/homo_sapiens/pep/
  - Sequence from UniProt: https://www.uniprot.org/uniprot/?query=+reviewed%3Ayes+AND+organism%3A%22Homo+sapiens+%28Human%29+%5B9606%5D%22&sort=score
- handle multiple UniProt cases
  - Try with bioMart
    - GRCh37: https://grch37.ensembl.org/biomart/martview/145410cb02e693b9da427f9c3dffe61f
    - GRCh38: http://uswest.ensembl.org/biomart/martview/84262bd39b9c7d095757b6ce493a2a9f
[...]

PR #501, opened 2021-04-06
fetch uniprot id in annotation summary

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/443

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/TranscriptConsequence.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/TranscriptConsequence.java
@@ -63,6 +63,7 @@ public class TranscriptConsequence
     private Double siftScore;
     private String siftPrediction;
     private String exon;
+    private String uniprotId;
 
     private List<String> refseqTranscriptIds;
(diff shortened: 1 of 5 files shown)


model/src/main/java/org/cbioportal/genome_nexus/model/TranscriptConsequence.java
```diff
@@ -63,6 +63,7 @@
     private Double siftScore;
     private String siftPrediction;
     private String exon;
+    private String uniprotId;
```
leexgh, line 66: Not sure if we want to have uniprot id in transcript consequence, because I saw we label transcript consequence with `@Deprecated` https://github.com/genome-nexus/genome-nexus/blob/master/model/src/main/java/org/cbioportal/genome_nexus/model/VariantAnnotationSummary.java#L17`

onursumer: I think it is okay to have `uniprotId` here. Seems like we only deprecate a specific old list here, not the actual model.

Issue #502, opened 2021-04-13
Query "uniprot_id" in "annotation_summary" is slow


leexgh:
We deployed v1.1.25 but have to reverted to v.1.1.24 because query "uniprot_id" in "annotation_summary" is slow. 

Reason: need to fetch data from two collections(`uniprot_id` is from `ensembl.biomart_transcript`, other fields in annotation_summary are from `vep.annotation`)

PR #510, opened 2021-04-14
Add transcript -> uniprot map into cache to increase query speed

leexgh:
Fix: #502 

Create a map: HashMap<transcript_id, uniprot_id>, store map in cache, and enrich annotation summary `uniprotId` with map.

Test results:

- Build transcript to uniprot map: 11s
- Query 1400 variant with map: 10s(no significant increase, same as before adding uniprot)


onursumer (approved):
LGTM

inodb:
@onursumer i kinda forgot how the canonical transcript implementation worked. That also uses some map right? But I guess the keys are only the canonical ensembl transcripts and the values the hugo symbols?

inodb (approved):
Looks pretty clean to me :) - thanks!

Issue #550, opened 2021-09-16
Genomic location normalization not working with annotation validating when follow up query is failing


leexgh:
For example:
```
7	55242467	55242486	AATTAAGAGAAGCAACATCT	AGCAA
```
There is a common prefix `A` in "**A**ATTAAGAGAAGCAACATCT" and "**A**GCAA", so genome nexus [normalizes](https://github.com/genome-nexus/genome-nexus/blob/master/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java#L118-L131) it to:
```
normalizedGenomicLocation = 
{
    "chromosome": "7",
    "start": 55242468,
    "end": 55242486,
    "referenceAllele": "ATTAAGAGAAGCAACATCT",
    "variantAllele": "GCAA"
}
```
Then generate hgvsg from normalizedGenomicLocation: 
```
7:g.55242468_55242486delinsGCAA
```
After getting VEP annotation we have an annotation validation to check responseReferenceAllele and providedReferenceAllele.
```
responseReferenceAllele: ATTAAGAGAAGCAACATCT
providedReferenceAllele: AATTAAGAGAAGCAACATCT
```
When they are not the same we [set VariantAllele to `-`](https://github.com/genome-nexus/genome-nexus/blob/9c185ce8f26345ce729fc191e28c0e1b788651fc/service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedGenomicLocationAnnotationServiceImpl.java#L161-L171) and send a follow up query, which would be:
```
followUpQueryGenomicLocation = 
{
    "chromosome": "7",
    "start": 55242467,
    "end": 55242486,
    "referenceAllele": "AATTAAGAGAAGCAACATCT",
    "variantAllele": "-"
}
```
From follow up query, follow_up_annotation is failed, so responseReferenceAllele = "-"
So on validating we have:
```
responseReferenceAllele: -
providedReferenceAllele: AATTAAGAGAAGCAACATCT
```
They are different so return annotation failure.

But in this case it should be able to return annotation.

PR #552, opened 2021-09-23
In follow up query, convert genomic location to hgvsg or region based on whether region annotation is enabled 

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/550

If region annotation is enabled: convert genomic location to region.
If region annotation is not enabled: convert genomic location to hgvs.

Test query:
7,55242467,55242486,AATTAAGAGAAGCAACATCT,AGCAA

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,4 +27,4 @@ services:
     ports:
       - "6060:8080"
     volumes:
-      - ${VEP_CACHE}:/opt/vep/.vep/:ro
+      - ${VEP_CACHE}:/opt/vep/.vep/
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/internal/GenomicLocationAnnotationServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/internal/GenomicLocationAnnotationServiceImpl.java
@@ -88,7 +88,7 @@ public class GenomicLocationAnnotationServiceImpl implements GenomicLocationAnno
     public VariantAnnotation getAnnotation(GenomicLocation genomicLocation)
         throws VariantAnnotationNotFoundException, VariantAnnotationWebServiceException
     {
-        VariantAnnotation variantAnnotation = this.variantAnnotationService.getAnnotation(this.notationConverter.genomicToEnsemblRestRegion(genomicLocation));
+        VariantAnnotation variantAnnotation = this.variantAnnotationService.getAnnotation(this.genomicLocationToVariantFormat.convert(genomicLocation));
         genomicLocation.setOriginalInput(genomicLocation.toString());
         variantAnnotation.setOriginalVariantQuery(genomicLocation.getOriginalInput());
         return variantAnnotation;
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -27,4 +27,4 @@ services:
     ports:
       - "6060:8080"
     volumes:
-      - ${VEP_CACHE}:/opt/vep/.vep/:ro
+      - ${VEP_CACHE}:/opt/vep/.vep/
```
leexgh, line 30: Need to remove "read-only", otherwise will have error:
```
java.nio.file.FileSystemException: /opt/vep/.vep/tmp: Read-only file system

at sun.nio.fs.UnixException.translateToIOException(UnixException.java:91)

at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)

at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)

at sun.nio.fs.UnixFileSystemProvider.createDirectory(UnixFileSystemProvider.java:384)

at java.nio.file.Files.createDirectory(Files.java:674)

at org.genomenexus.vep_wrapper.VepRunner.createTmpDirIfNecessary(VepRunner.java:62)

at org.genomenexus.vep_wrapper.VepRunner.run(VepRunner.java:119)

at org.genomenexus.vep_wrapper.VepController.getVepAnnotation(VepController.java:47)
```

service/src/main/java/org/cbioportal/genome_nexus/service/internal/GenomicLocationAnnotationServiceImpl.java
```diff
@@ -88,7 +88,7 @@ public GenomicLocationAnnotationServiceImpl(CachedVariantRegionAnnotationFetcher
     public VariantAnnotation getAnnotation(GenomicLocation genomicLocation)
         throws VariantAnnotationNotFoundException, VariantAnnotationWebServiceException
     {
-        VariantAnnotation variantAnnotation = this.variantAnnotationService.getAnnotation(this.notationConverter.genomicToEnsemblRestRegion(genomicLocation));
+        VariantAnnotation variantAnnotation = this.variantAnnotationService.getAnnotation(this.genomicLocationToVariantFormat.convert(genomicLocation));
```
sheridancbio, line 91: Yes, this is good. The concrete instance of the functional interface for genomicLocationToVariantFormat is set according to isRegionAnnotationEnabled(). When we have an attached vep (from genome-nexus-vep) we send annotation requests in RestRegion format to VEP - so genomicLocationToVariantFormat.convert() will be set to notationConverter::genomicToEnsemblRestRegion. Otherwise, we will be sending requests (in Hgvs format) to the public Ensembl VEP API - so so genomicLocationToVariantFormat.convert() will be set to notationConverter:: genomicToHgvs.
This change fixes a bug - where we were always converting the genomicLocation to RestRegion no matter what configuration we were in. However ... seeing this bug now, I don't see how any request could succeed when sent to the POST endpoint with a genomicLocation object supplied by the user. It seems like those should also be improperly converted to RestRegion and then fail annotation in the Ensembl Rest API (which expect hgvs). But this request succeeds on the current public genome nexus:
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' -d '[ \ 
   { \ 
     "chromosome": "7", \ 
     "start": 55242468, \ 
     "end": 55242469, \ 
     "referenceAllele": "AT", \ 
     "variantAllele": "GC" \ 
   } \ 
 ]' 'https://www.genomenexus.org/annotation/genomic'
Shouldn't this query also (with the current code) be converted to Ensembl Rest Region and then be sent to the Public Ensembl API and fail? I guess this endpoint actually is supplying a *list* of GenomicLocations, so the call is to getAnnotations(List<GenomicLocation>, but that function (below) also has the bug. AHA .. it calls the function with the extra arguments (genomicLocations, isoformOverrideSource, token, and fields) and that function does not have the bug. Ok, it makes sense now.

sheridancbio (approved):
Nice work here 👍

sheridancbio:
code tested on annotation deployment of genome-nexus

PR #580, opened 2021-12-22
Explicitly set log4j2 version

### LINKED ISSUE #579: Make sure to use the latest version of log4j in backend
Currently `org.apache.commons.logging.LogFactory` creates a LOG4J logger by default. We should make sure that log4j dependency is not outdated in GenomeNexus backend.

onursumer:
Fix #579

See https://spring.io/blog/2021/12/10/log4j2-vulnerability-and-spring-boot for details.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,6 +39,7 @@
 
     <properties>
       <java.version>1.8</java.version>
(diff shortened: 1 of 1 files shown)


leexgh (approved):
👍

PR #587, opened 2022-02-03
Fix 404s

inodb:
Used to give 404 on frontend only pages:

```
curl --verbose 'http://localhost:38080/variant/7:g.140453136A%3ET'
< HTTP/1.1 404
```

This fixes it to instead show 200:
```
curl --verbose 'http://localhost:38080/variant/7:g.140453136A%3ET'
< HTTP/1.1 200
```

However this means 404s should now be handled in the frontend.
See for instance:

```
curl --verbose 'http://localhost:38080/api/random'
< HTTP/1.1 200
```

For cBioPortal's API it seems that `/api/error` is handled by API but other routes like `/asddasd/` are handled by frontend, which might be a cleaner solution.


leexgh (approved):
Thanks for fixing!

PR #608, opened 2022-05-03
Update My variant info model

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/610
There is a new field `MyVariantInfo.ClinVar.Hgvs.Protein`


inodb (approved):
LGTM - thank you!!

Issue #609, opened 2022-05-03
TNP is annotated as ONP


inodb:
This variant:

https://www.genomenexus.org/annotation/13:g.48954120_48954122delinsAAA?fields=hotspots,annotation_summary,my_variant_info,clinvar,signal

According to the MAF format this should be a TNP:

>  Type of mutation. TNP (tri-nucleotide polymorphism) is analogous to DNP (di-nucleotide polymorphism) but for three consecutive nucleotides. ONP (oligo-nucleotide polymorphism) is analogous to TNP but for consecutive runs of four or more (SNP, DNP, TNP, ONP, INS, DEL, or Consolidated)

But it gets annotated as ONP:

```
"variantType": "ONP",
```

Did we ever annotate TNPs correctly?

Let's fix the logic here:

https://github.com/genome-nexus/genome-nexus/blob/fc06c71c65c7bd3b4febd224ff36bb77a361d4a4/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/VariantTypeResolver.java#L51

and add a test to make sure it continues to work

PR #613, opened 2022-05-10
Fix TNP variant type

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/609


inodb (approved):
LGTM 👍 - nice test cases as well!

onursumer (approved):
🙌

sheridancbio (commented):
This seems ok to me, but I do wonder why we were originally returning ONP for 3 NT variants. Maybe @ao508 has something to say (I think I have a faint memory from a few years back about this).

I did search through the genome-nexus-annotation-pipeline code for any references to DNP/TNP/ONP but found nothing there which would suggest against adopting this change.

PR #619, opened 2022-06-02
Version Support

averyniceday:
Changes include
- New models (3 levels including version, source-specific info, aggregate info returned by endpoint)
- Repository layer (autowires all info related to versioning)
- New properties to specify versions

Currently supports versioning for Genome Nexus (server/codebase and mongo) and VEP (server/codebase and cache)

Example in Swagger
```
{
  "genomeNexus": {
    "database": {
      "static": true,
      "version": "string"
    },
    "server": {
      "static": true,
      "version": "string"
    }
  },
  "vep": {
    "cache": {
      "static": true,
      "version": "string"
    },
    "comment": "string",
    "server": {
      "static": true,
      "version": "string"
    }
  }
}
```

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/VEPInfo.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/VEPInfo.java
@@ -1,0 +1,25 @@
+package org.cbioportal.genome_nexus.model;
+
+public class VEPInfo {
+
+    private final String NA_STRING = "NA";
+    private final String COMMENT = "VEP annotations are currently externally sourced from ENSEMBL. Results are subject to change without notice.";
+
+    // server corresponds to genome nexus VEP code release
+    // cache corresponds to downloaded FASTA cache
+    // comment is an additional disclaimer in the event we are using ENSEMBL VEP
+    public Version server;
+    public Version cache;
+    public String comment;
+
+    public VEPInfo(String vepServerVersion, String vepCacheVersion, String vepRegionURL) {
+        if (vepRegionURL != null && !vepRegionURL.isEmpty()) {
+            this.server = new Version(vepServerVersion, true);
+            this.cache = new Version(vepCacheVersion, true);
+        } else {
+            this.server = new Version(NA_STRING, false);
+            this.cache = new Version(NA_STRING, false);
+            this.comment = COMMENT;
+        }
+    }
+}
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/Version.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/Version.java
@@ -1,0 +1,28 @@
+package org.cbioportal.genome_nexus.model;
+
+public class Version {
+
+    private String version;
+    private boolean isStatic;
+
+    public Version(String version, boolean isStatic) {
+        this.version = version;
+        this.isStatic = isStatic;
+    }
+
+    public String getVersion() {
+        return version;
+    }
+
+    public void setVersion(String version) {
+        this.version = version;
+    }
+
+    public boolean isStatic() {
+        return isStatic;
+    }
+
+    public void setStatic(boolean isStatic) {
+        this.isStatic = isStatic;
+    }
+}
--- a/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/SourceInfoRepository.java
+++ b/persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/SourceInfoRepository.java
@@ -1,0 +1,41 @@
+package org.cbioportal.genome_nexus.persistence.internal;
+
+import org.cbioportal.genome_nexus.model.AggregateSourceInfo;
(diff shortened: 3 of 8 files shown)


model/src/main/java/org/cbioportal/genome_nexus/model/VEPInfo.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class VEPInfo {
+
+    private final String NA_STRING = "NA";
+    private final String COMMENT = "VEP annotations are currently externally sourced from ENSEMBL. Results are subject to change without notice.";
+
+    // server corresponds to genome nexus VEP code release
+    // cache corresponds to downloaded FASTA cache
+    // comment is an additional disclaimer in the event we are using ENSEMBL VEP
+    public Version server;
```
sheridancbio, line 11: The more standard form would be private data members with get()/set() accessor functions. (Comment applies to GenomeNexusInfo.java too). With this form, we may see (in other classes' member functions) things like "returnedVepInfo.cache = updatedVersion;" instead of "returnedVepInfo.setCache(updatedVersion);"
This is not really a big deal unless at some point down the road we want to alter the implementation of VEPInfo to something with a different data representation (which is unlikely)

model/src/main/java/org/cbioportal/genome_nexus/model/VEPInfo.java
```diff
@@ -0,0 +1,25 @@
...
+    public Version cache;
+    public String comment;
+
+    public VEPInfo(String vepServerVersion, String vepCacheVersion, String vepRegionURL) {
```
sheridancbio, line 15: By defining only a complex (non-default) constructor, it means that it will not be possible to construct a VEPInfo object without knowing the argument values for all three parameters here. So creating temporary / placeholder objects which get fully populated later will not be possible. Again, probably not a problem for this simple type which has only a single purpose. But in general, my concept of the "ideal" model class is a POJO with no business logic, containing only private data members, getters and setters which do direct returns / assignments, and a default constructor.

If we were to use kind of model, this business logic in this method would move into the service layer implementation, or into a service layer utility/component like "VEPInfoFactory.constructVEPInfo(String vepServerVersion, String vepCacheVersion, String vepRegionURL)". I'm not requesting that change ... just sharing the picture. One advantage of that approach is to make unit test writing easier, although in this case no unit testing is needed.

model/src/main/java/org/cbioportal/genome_nexus/model/Version.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public String getVersion() {
+        return version;
+    }
+
+    public void setVersion(String version) {
+        this.version = version;
+    }
+
+    public boolean isStatic() {
```
sheridancbio, line 21: I think somewhere we should have a description (to the end user) of what Static means. There is not really a good single word which captures what we mean by this boolean ... something like "expected to deliver unchanging results for a given request until the string accessed through getVersion() changes"

persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/SourceInfoRepository.java
```diff
@@ -0,0 +1,41 @@
...
+    @Autowired
+    private GenomeNexusInfo genomeNexusInfo;
+
+    @Autowired
+    private VEPInfo vepInfo;
+
+    @Autowired
+    private AggregateSourceInfo aggregateSourceInfo;
+
+    @Bean
```
sheridancbio, line 23: good .. as beans these will only be initialized a single time during startup and will then be reused by reference throughout the server execution.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/InfoServiceImpl.java
```diff
@@ -1,36 +1,24 @@
...
-            this.version = version;
-        }
+    @Autowired
+    private SourceInfoRepository sourceInfoRepository;
 
-        public String getVersion() {
-            return this.version;
-        }
+    public AggregateSourceInfo getAggregateSourceInfo() {
+        return sourceInfoRepository.getAggregateSourceInfo();
```
sheridancbio, line 21: 👍

sheridancbio (approved):
looks good. I shared some thoughts.

web/src/main/resources/application.properties.EXAMPLE
```diff
@@ -14,6 +14,8 @@ vep.max_page_size=200
 # This is only the VEP part of the Ensembl REST API.
 # e.g. when running on port 6060:
 # gn_vep.region.url=http://localhost:6060/vep/human/region/VARIANT
+# gn_vep.cache.version=
+# gn_vep.server.version=
```
pieterlukasse, line 18: @averyniceday @sheridancbio: it would be nice to add some comments on when and why to set these properties.

(we are currently going through the latest PRs to understand what has changed)

PR #620, opened 2022-07-07
Return separate annotations for identical variant in different representations

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/9517
See comment: https://github.com/cBioPortal/cbioportal/issues/9517#issuecomment-1175370011
For example:
```
[
  {
    "chromosome": "12",
    "start": 25398284,
    "end": 25398284,
    "referenceAllele": "C",
    "variantAllele": "A"
  },
  {
    "chromosome": "12",
    "start": 25398283,
    "end": 25398284,
    "referenceAllele": "AC",
    "variantAllele": "AA"
  }
]
```
This example case is from [GENIE](https://genie.cbioportal.org/results/mutations?cancer_study_list=genie_public&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&profileFilter=mutations&case_set_id=genie_public_all&gene_list=KRAS&geneset_list=%20&tab_index=tab_visualize&Action=Submit) 
Ideally the second variant should be normalized to the first variant, so technically they are identical. But if for any reason users query these two variants, genome nexus should return annotation for both variants. Because otherwise we probably have some variants that couldn't find a matching annotation in return, so the number of input variant should equal to number of output annotation. In this example, two annotations should be the same, but one has `"originalVariantQuery": "12,25398284,25398284,C,A"`, the other has `"originalVariantQuery": "12,25398283,25398284,AC,AA"`. 

#### Input
```
[
  {
    "chromosome": "12",
    "start": 25398284,
    "end": 25398284,
    "referenceAllele": "C",
    "variantAllele": "A"
  },
  {
    "chromosome": "12",
    "start": 25398283,
    "end": 25398284,
    "referenceAllele": "AC",
    "variantAllele": "AA"
  }
]
```
#### Before
Example return:
```
[
  {
    "variant": "12:g.25398284C>A",
    ...
    ...
    "originalVariantQuery": "12,25398283,25398284,AC,AA",
  }
]
```

#### After
Example return:
```
[
  {
    "variant": "12:g.25398284C>A",
    ...
    ...
    "originalVariantQuery": "12,25398284,25398284,C,A",
  },
  {
    "variant": "12:g.25398284C>A",
    ...
    ...
    "originalVariantQuery": "12,25398283,25398284,AC,AA",
  }
]
```

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/internal/GenomicLocationAnnotationServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/internal/GenomicLocationAnnotationServiceImpl.java
@@ -43,6 +43,7 @@ import org.cbioportal.genome_nexus.service.exception.VariantAnnotationNotFoundEx
 import org.cbioportal.genome_nexus.service.exception.VariantAnnotationWebServiceException;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.stereotype.Service;
+import com.google.gson.Gson;
 import org.springframework.beans.factory.annotation.*;
 
 import java.util.*;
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/internal/GenomicLocationAnnotationServiceImpl.java
```diff
@@ -141,28 +154,42 @@ public List<VariantAnnotation> getAnnotations(List<GenomicLocation> genomicLocat
...
+                variantAnnotations.add(a);
+            }
+            else {
+                for (String origVarQuery : convertedVarsToOrigVarQueryMap.get(a.getVariant())) {
+                    // if query multiple "identical" variants
+                    // for example "12,25398284,25398284,C,A" and "12,25398283,25398284,AC,AA"
+                    // both variants should be returned
+                    Gson gson = new Gson();
+                    VariantAnnotation tempAnnotation = gson.fromJson(gson.toJson(a), VariantAnnotation.class);
+                    tempAnnotation.setOriginalVariantQuery(origVarQuery);
```
leexgh, line 177: @onursumer Any better way to do deep copy?

onursumer (approved):
LGTM

inodb (approved):
Thank you!

Issue #621, opened 2022-07-07
How to make the hotspots endpoint to use the local VEP service?


ruslan-forostianov:
The hotspots endpoint (`/cancer_hotspots/genomic`) fetches data with `vep.url` url.

How to make the hotspots take advantage of the running local VEP service (https://github.com/genome-nexus/genome-nexus-vep)?
`gn_vep.region.url` seems to be ignored by the hotspots endpoint.

PR #622, opened 2022-07-07
Add a possibility to use local vep for hotspots endpoint

ruslan-forostianov:
Issue: https://github.com/genome-nexus/genome-nexus/issues/621

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/GenomicLocationAnnotationService.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/GenomicLocationAnnotationService.java
@@ -36,9 +36,11 @@ import org.cbioportal.genome_nexus.model.GenomicLocation;
 import org.cbioportal.genome_nexus.model.VariantAnnotation;
 import org.cbioportal.genome_nexus.service.exception.VariantAnnotationNotFoundException;
 import org.cbioportal.genome_nexus.service.exception.VariantAnnotationWebServiceException;
+import org.cbioportal.genome_nexus.service.internal.GenomicLocationAnnotationServiceImpl;
 
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 
 public interface GenomicLocationAnnotationService
 {
@@ -50,4 +52,6 @@ public interface GenomicLocationAnnotationService
     VariantAnnotation getAnnotation(String genomicLocation, String isoformOverrideSource, Map<String, String> token, List<String> fields)
         throws VariantAnnotationWebServiceException, VariantAnnotationNotFoundException;
     List<VariantAnnotation> getAnnotations(List<GenomicLocation> genomicLocations, String isoformOverrideSource, Map<String, String> token, List<String> fields);
+
+    Function<GenomicLocation, String> getGenomicLocationToVariantFormat();
 }
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/GenomicLocationAnnotationService.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/GenomicLocationAnnotationService.java
@@ -36,7 +36,6 @@ import org.cbioportal.genome_nexus.model.GenomicLocation;
 import org.cbioportal.genome_nexus.model.VariantAnnotation;
 import org.cbioportal.genome_nexus.service.exception.VariantAnnotationNotFoundException;
 import org.cbioportal.genome_nexus.service.exception.VariantAnnotationWebServiceException;
-import org.cbioportal.genome_nexus.service.internal.GenomicLocationAnnotationServiceImpl;
 
 import java.util.List;
 import java.util.Map;
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/GenomicLocationAnnotationService.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/GenomicLocationAnnotationService.java
@@ -39,7 +39,6 @@ import org.cbioportal.genome_nexus.service.exception.VariantAnnotationWebService
 
 import java.util.List;
 import java.util.Map;
-import java.util.function.Function;
 
 public interface GenomicLocationAnnotationService
 {
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/GenomicLocationAnnotationService.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/GenomicLocationAnnotationService.java
@@ -52,5 +52,5 @@ public interface GenomicLocationAnnotationService
         throws VariantAnnotationWebServiceException, VariantAnnotationNotFoundException;
     List<VariantAnnotation> getAnnotations(List<GenomicLocation> genomicLocations, String isoformOverrideSource, Map<String, String> token, List<String> fields);
 
-    Function<GenomicLocation, String> getGenomicLocationToVariantFormat();
+    String getVariantFormat(GenomicLocation genomicLocation);
 }
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
@@ -36,8 +36,8 @@ import org.cbioportal.genome_nexus.component.annotation.HotspotFilter;
 import org.cbioportal.genome_nexus.model.*;
 import org.cbioportal.genome_nexus.persistence.HotspotRepository;
 import org.cbioportal.genome_nexus.service.CancerHotspotService;
+import org.cbioportal.genome_nexus.service.GenomicLocationAnnotationService;
 import org.cbioportal.genome_nexus.service.VariantAnnotationService;
-import org.cbioportal.genome_nexus.component.annotation.NotationConverter;
 import org.cbioportal.genome_nexus.service.exception.CancerHotspotsWebServiceException;
 import org.cbioportal.genome_nexus.service.exception.VariantAnnotationNotFoundException;
 import org.cbioportal.genome_nexus.service.exception.VariantAnnotationWebServiceException;
@@ -45,6 +45,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import java.util.*;
+import java.util.function.Function;
 
 /**
  * @author Selcuk Onur Sumer
@@ -52,21 +53,23 @@ import java.util.*;
 @Service
 public class CancerHotspotServiceImpl implements CancerHotspotService
 {
+
     private final HotspotRepository hotspotRepository;
     private final VariantAnnotationService variantAnnotationService;
     private final HotspotFilter hotspotFilter;
-    private final NotationConverter notationConverter;
+
+    private final GenomicLocationAnnotationService genomicLocationAnnotationService;
 
     @Autowired
     public CancerHotspotServiceImpl(HotspotRepository hotspotRepository,
                                     VariantAnnotationService verifiedHgvsVariantAnnotationService,
-                                    HotspotFilter hotspotFilter,
-                                    NotationConverter notationConverter)
+                                    GenomicLocationAnnotationService verifiedGenomicLocationAnnotationServiceImpl,
+                                    HotspotFilter hotspotFilter)
     {
         this.hotspotRepository = hotspotRepository;
         this.variantAnnotationService = verifiedHgvsVariantAnnotationService;
+        this.genomicLocationAnnotationService = verifiedGenomicLocationAnnotationServiceImpl;
         this.hotspotFilter = hotspotFilter;
-        this.notationConverter = notationConverter;
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
```diff
@@ -156,25 +159,29 @@ public List<Hotspot> getHotspotAnnotationsByGenomicLocation(String genomicLocati
...
+        VariantAnnotation variantAnnotation = genomicLocationAnnotationService.getAnnotation(genomicLocation);
+        return this.getHotspotAnnotations(variantAnnotation);
     }
 
     @Override
     public List<AggregatedHotspots> getHotspotAnnotationsByGenomicLocations(List<GenomicLocation> genomicLocations)
         throws CancerHotspotsWebServiceException
     {
-        // convert genomic location to hgvs notation (there is always 1-1 mapping)
-        Map<String, GenomicLocation> variantToGenomicLocation = notationConverter.genomicToHgvsMap(genomicLocations);
```
pieterlukasse, line 169: what is the reason we can't use this function anymore? Does it make external calls as well?

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
```diff
@@ -156,25 +159,29 @@ public List<Hotspot> getHotspotAnnotationsByGenomicLocation(String genomicLocati
         throws VariantAnnotationNotFoundException, VariantAnnotationWebServiceException,
         CancerHotspotsWebServiceException
     {
-        GenomicLocation location = this.notationConverter.parseGenomicLocation(genomicLocation);
-
-        return this.getHotspotAnnotationsByVariant(this.notationConverter.genomicToHgvs(location));
+        VariantAnnotation variantAnnotation = genomicLocationAnnotationService.getAnnotation(genomicLocation);
```
pieterlukasse, line 162: Nice 👍 

This is basically doing:
```java
    VariantAnnotation variantAnnotation = genomicLocationAnnotationService.getAnnotation(genomicLocation);
```
instead of
```java
    VariantAnnotation variantAnnotation = variantAnnotationService.getAnnotation(this.notationConverter.genomicToHgvs(this.notationConverter.parseGenomicLocation(genomicLocation)))
```

I think the remaining question is: do we know if these can be expected to return (mostly) the same results? How can we test that?

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
ruslan-forostianov: no, it does not make external calls. But it only works with HgvsVariantAnnotationService (no `gn_vep.region.url` set). When we use `RegionVariantAnnotationService`, we need a map from ensembl rest region to genomic location instead (`notationConverter::genomicToEnsemblRestRegion`, not `notationConverter::genomicToHgvs`).

But after that said, we only need this map to trace back and set a genomic location for each result entry.

There must be better way. e.g. a variant annotation already has all necessary information (chromosome, start, end, ref and variant allele) to construct a genomic location instance.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
ruslan-forostianov: I think it depends what version of vep cache you have locally vs what version of data we get via `vep.url` endpoint. The first version of the call can be configured to fetch data whether locally or from the ensembl endpoint. The second only from the ensembl endpoint.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
```diff
@@ -156,25 +159,29 @@ public List<Hotspot> getHotspotAnnotationsByGenomicLocation(String genomicLocati
...
+        Map<String, GenomicLocation> variantToGenomicLocation = new HashMap<>();
+        for (GenomicLocation genomicLocation: genomicLocations) {
+            variantToGenomicLocation.put(genomicLocationToVariantFormat.apply(genomicLocation), genomicLocation);
+        }
+        List<AggregatedHotspots> hotspots = new ArrayList<>();
+        for (VariantAnnotation variantAnnotation : variantAnnotations)
+        {
+            AggregatedHotspots aggregatedHotspots = new AggregatedHotspots();
+            aggregatedHotspots.setHotspots(this.getHotspotAnnotations(variantAnnotation));
+            aggregatedHotspots.setVariant(variantAnnotation.getVariant());
```
pieterlukasse, line 182: if `getVariant()` is different in this case, we might be breaking something. Let's check this.

ruslan-forostianov:
@sheridancbio Thanks for great feedback!

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
ruslan-forostianov: @sheridancbio yes, I agree. Having one internal format (a custom class, not a string) and converting and serialising it to another required format of the external call will eliminate quite some code duplication.

Is there an open proposal for that?

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
ruslan-forostianov: I've checked that the variant field does not seem to be used anywhere really. Maybe it's worth removing altogether?
@sheridancbio Is genomenexus api used by cbioportal only? What'd be a procedure to introduce a non-backward compatible change?

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
```diff
@@ -156,25 +159,29 @@ public List<Hotspot> getHotspotAnnotationsByGenomicLocation(String genomicLocati
...
-        List<AggregatedHotspots> hotspots = this.getHotspotAnnotationsByVariants(
-            new ArrayList<>(variantToGenomicLocation.keySet()));
-
-        // add genomic location info too
-        for (AggregatedHotspots aggregatedHotspots: hotspots) {
-            aggregatedHotspots.setGenomicLocation(variantToGenomicLocation.get(aggregatedHotspots.getVariant()));
+        Function<GenomicLocation, String> genomicLocationToVariantFormat = this.genomicLocationAnnotationService.getGenomicLocationToVariantFormat();
+        Map<String, GenomicLocation> variantToGenomicLocation = new HashMap<>();
+        for (GenomicLocation genomicLocation: genomicLocations) {
+            variantToGenomicLocation.put(genomicLocationToVariantFormat.apply(genomicLocation), genomicLocation);
```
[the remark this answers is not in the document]
ruslan-forostianov: yes, makes sense.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
ruslan-forostianov: The swagger says that the api is "internal" on this page https://www.genomenexus.org/swagger-ui.html

I am not sure what exactly it means in this context.

ruslan-forostianov:
@sheridancbio I've just pushed the improvements to the code. Could you please have another look?

service/src/main/java/org/cbioportal/genome_nexus/service/GenomicLocationAnnotationService.java
```diff
@@ -39,6 +39,7 @@
 
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
```
leexgh, line 42: Maybe we don't need to import this? And same for other files below

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
```diff
@@ -36,37 +36,40 @@
...
-    private final NotationConverter notationConverter;
+
+    private final GenomicLocationAnnotationService genomicLocationAnnotationService;
 
     @Autowired
     public CancerHotspotServiceImpl(HotspotRepository hotspotRepository,
                                     VariantAnnotationService verifiedHgvsVariantAnnotationService,
-                                    HotspotFilter hotspotFilter,
-                                    NotationConverter notationConverter)
+                                    GenomicLocationAnnotationService verifiedGenomicLocationAnnotationServiceImpl,
```
leexgh, line 66: To keep the consistency, we can rename this to `verifiedGenomicLocationAnnotationService`

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
```diff
@@ -156,25 +159,28 @@ public List<Hotspot> getHotspotAnnotationsByGenomicLocation(String genomicLocati
...
-        List<AggregatedHotspots> hotspots = this.getHotspotAnnotationsByVariants(
-            new ArrayList<>(variantToGenomicLocation.keySet()));
-
-        // add genomic location info too
-        for (AggregatedHotspots aggregatedHotspots: hotspots) {
-            aggregatedHotspots.setGenomicLocation(variantToGenomicLocation.get(aggregatedHotspots.getVariant()));
+        Map<String, GenomicLocation> variantToGenomicLocation = new HashMap<>();
+        for (GenomicLocation genomicLocation: genomicLocations) {
+            variantToGenomicLocation.put(genomicLocationAnnotationService.getVariantFormat(genomicLocation), genomicLocation);
+        }
```
leexgh, line 175: We can use stream to simplify this, something like
```suggestion
Map<String, GenomicLocation> variantToGenomicLocation = genomicLocations.stream()
    .collect(Collectors.toMap(
        genomicLocation -> genomicLocationAnnotationService.getVariantFormat(genomicLocation),
        v -> v)
    );
```

service/src/main/java/org/cbioportal/genome_nexus/service/GenomicLocationAnnotationService.java
ruslan-forostianov: @leexgh Xiang, thanks! I'm going to optimise imports.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
ruslan-forostianov: @leexgh it's not gonna work. There are 2 implementations of `GenomicLocationAnnotationService` interface: `GenomicLocationAnnotationServiceImpl` and `VerifiedGenomicLocationAnnotationServiceImpl`. Spring's `@Autowired` falls back on selecting an implementation based on exact naming when there are multiple implementation of the type.

ruslan-forostianov:
@leexgh Hey Xiang, I've resolved your improvements points. Could you please have another look?

leexgh (approved):
Looks good to me. Thanks!

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
sheridancbio: I guess I'm just sharing a concept rather than actually proposing the code be refactored now to align with that concept (which would be a big job). There is probably more helpful refactoring which could be done, so I don't think this concept should be prioritized. But if an opportunity comes up (such as if we were to do a major refactoring), then maybe it should be considered.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/CancerHotspotServiceImpl.java
sheridancbio: @ruslan-forostianov The two main uses of the genomenexus api (in particular the annotation endpoint) are:
1. the cBioPortal website uses it for dynamic data requests during the rendering of certain pages (such as the mutation viewer / lollypop display)
2. the MSK importer codebase, and related batch annotation tools used by the data curators call the genomenexus api to annotate genomic events. We try to "pre-annotate" our files so that loading occurs faster. We also have a nightly fetch pipeline which annotates  the events for the samples added to the MSK cBioPortal site, and that fetch pipeline uses genome-nexus. But if an un-annotated mutations file gets imported, the the annotation will be done at time of import. These calls all go through the genome nexus client library in https://github.com/genome-nexus/genome-nexus-annotation-pipeline

There may be other institutions which use the genome-nexus api, but I am not aware of a specific case. We did introduce changes which were (I think) not backwards compatible a few years back. I think I remember that we added a parameter to annotation requests which was a maximum execution time (after which the job would auto-cancel). This might have been in the genome-nexus-vep codebase however. We added code to always report the original query string when responding to a request for annotation, I think. (although maybe adding a field to the response still could be considered backwards compatible). But anyway, I believe we are free to make adjustments to the API when it makes sense. cBioPortal (and related importers) must be the dominant user of the system. My opinion is that API-breaking changes could be made, so long as a transition plan for rolling them out is agreed on by all the main clients. However, there are many installations of cBioPortal outside of the core development teams, so if the change would disrupt function of the cBioPortal frontend codebase we would probably need to version the API so that the older API would still be available. Other developers of genome-nexus should comment on this too.

Issue #627, opened 2022-08-17
Optimize mutation assessor performance


leexgh:
Mutation Assessor API is down so we are importing mutation assessor data into genome nexus database. We need to do some improvements for mutation assessor queries, because the original implementation was designed for fetching API and cache to database, now we switch to database queries so there are some optimization could do.
Tasks:
- [x] Add index to mutation assessor
- [x] Remove all external mutation assessor fetching methods 
- [x] Replace mutation assessor tests with db query rather than external query

PR #628, opened 2022-08-18
Adding index to optimize mutation assessor performance

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/627

Test instance: https://beta.genomenexus.org/
Add 'mutation_assessor' in fileds


inodb (approved):
I like it 🏆! Thanks so much @leexgh ! Great work!

pieterlukasse:
This is great 👍 
Less external calls, better data privacy for local installations.

PR #633, opened 2022-09-19
Add dockerfile and annotation.

price0416:
Added Dockerfile to the /container/ directory.  Updated /docs/Annotate-File.md to include documentation for working with this docker image.

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -1,0 +1,100 @@
+FROM ubuntu:18.04
+
+LABEL maintainer="Adam Price (pricea2@mskcc.org)" \
+      contributor="Anne Marie Noronha (noronhaa@mskcc.org)" \
+      version.image="0.0.1" \
+      version.annotation-tools="1.0.0" \
+      version.annotation-pipeline="1.0.0" \
+      version.python="3.9.7" \
+      version.java="1.8.0" \
+      version.maven="3.8.4" \
+      source.annotation-tools="https://github.com/genome-nexus/annotation-tools" \
+      source.annotation-pipeline="https://github.com/genome-nexus/genome-nexus-annotation-pipeline" \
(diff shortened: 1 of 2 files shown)


pieterlukasse:
@price0416 @inodb thanks for the PR and making the dependency issue clear. 
Here are my thoughts on this PR:
- the current repository already has a Dockerfile. The current Dockerfile correctly encapsulates the service implemented in this repository
   - the Dockerfile you propose should probably get its own repository. I think it could cause confusion if it stays in this repository, given the other Dockerfile. The new repository could be called something like genome-nexus/all-in-one or something like that...
- instead of adding all .sh scripts to the docker image, another way to do it could be to have relevant wrapper .sh scripts outside of the image, just calling `docker run` commands for the various steps that are already present in the docker images of the different parts (please note that https://github.com/genome-nexus/annotation-tools/pull/49 will add a Dockerfile for annotation-tools).

Let me know what you think! We can also connect in slack to discuss in more detail if that helps!

pieterlukasse (commented):
@price0416 @inodb thanks for the PR and making the dependency issue clear. 
Here are my thoughts on this PR:
- the current repository already has a Dockerfile. The current Dockerfile correctly encapsulates the service implemented in this repository
   - the Dockerfile you propose should probably get its own repository. I think it could cause confusion if it stays in this repository, given the other Dockerfile. The new repository could be called something like genome-nexus/all-in-one or something like that...
- instead of adding all .sh scripts to the docker image, another way to do it could be to have relevant wrapper .sh scripts outside of the image, just calling `docker run` commands for the various steps that are already present in the docker images of the different parts (please note that https://github.com/genome-nexus/annotation-tools/pull/49 will add a Dockerfile for annotation-tools).

Let me know what you think! We can also connect in slack to discuss in more detail if that helps!

inodb:
Thanks @pieterlukasse! That makes sense to me

Issue #644, opened 2022-12-02
`/annotation/genomic` returns variant annotations with empty colocated variants


ruslan-forostianov:
I've found the bug while using the `genome-nexus-annotation-pipeline`. I've noticed that `dbSNP_RS` column of maf file were not populated consistently. Sometimes I would get this column blank, while most of the times for the same genomic coordinates it was populated.

I dig deeper and found out that `genome-nexus` does not populate colocated variants (used to get `dbSNP_RS`) when asked about the genomic location that is not yet cached. However, It does populate colocated variants when reads result from the cache (mongodb).

ruslan-forostianov:
Whether you get colocated variants in the result or not depends on the following condition https://github.com/genome-nexus/genome-nexus/blob/6841e5c49f89f1f8b399b524fb10be6d31cbbf97/service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java#L85

ruslan-forostianov:
Given [the current VEP output](https://www.ensembl.org/info/docs/tools/vep/vep_formats.html#json), it does not seem like the rest of the fields of `ColocatedVariant` (like `gnomad_nfe_maf`, `gnomad_nfe_allele`) get ever populated. Fields with similar names could be found under `"frequencies"` field though.

ruslan-forostianov:
We might consider to parse `VariantAnnotation` from the VEP output and save the bean to the mongodb cache.  At the moment, we save the raw VEP output including fields we never use to the mongodb cache. https://github.com/genome-nexus/genome-nexus/blob/6841e5c49f89f1f8b399b524fb10be6d31cbbf97/service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java#L103

PR #647, opened 2022-12-07
Deserialise VariantAnnotation.colocatedVariants on first hit

ruslan-forostianov:
Fix filling of the colocated variants field when the variant annotation is not in the cache yet. e.g. fetched for such genomic locations for the first time.

The issue is described in more details here https://github.com/genome-nexus/genome-nexus/issues/644


pieterlukasse (approved):
great fix 👍

inodb (approved):
LGTM - thanks!

PR #671, opened 2023-03-06
Reading VUEs.json from resources folder

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/672
- Tests are failing because it can't read relative path of VUEs.json in `data` folder. Change to reading `VUEs.json` from `resources` instead.
- Also fix circle ci to make it run on all pull requests


inodb (approved):
Great! Thanks for fixing!

Issue #642, opened 2022-11-17
Annotation failed 10,89720811,89720811,C,CA 


leexgh:
`10,89720811,89720811,C,CA` annotation fails but `10,89720811,89720811,-,A` and `10:g.89720811_89720812insA` can get annotation.
https://www.genomenexus.org/annotation/genomic/10,89720811,89720811,C,CA?fields=annotation_summary
https://www.genomenexus.org/annotation/genomic/10,89720811,89720811,-,A?fields=annotation_summary
https://www.genomenexus.org/annotation/10%3Ag.89720811_89720812insA?fields=annotation_summary

PR #673, opened 2023-03-13
Fix failing insertion annotation with matching alleles

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/642

`10,89720811,89720811,C,CA` annotation fails but `10,89720811,89720811,-,A` and `10:g.89720811_89720812insA` can get annotation.
This is because the annotation didn't pass `verifyOrFailAnnotation`. 

Changes in this PR:
Change`verifyOrFailAnnotation` to be able to verify insertion with matching alleles


service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedGenomicLocationAnnotationServiceImpl.java
```diff
@@ -139,8 +139,8 @@ private VariantAnnotation verifyOrFailAnnotation(VariantAnnotation annotation)
         }
         LOG.debug("verifying providedReferenceAllele : '" + providedReferenceAllele + "'");
         String responseReferenceAllele = getReferenceAlleleFromAnnotation(annotation);
-        if (responseReferenceAllele.length() != providedReferenceAllele.length()) {
-            // for altered length Deletion-Insertion responses, recover full reference allele with followup query
```
leexgh, line 143: Any reason it's for `Deletion-Insertion` only?

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedGenomicLocationAnnotationServiceImpl.java
```diff
@@ -161,14 +161,31 @@ private VariantAnnotation verifyOrFailAnnotation(VariantAnnotation annotation)
...
+        // for altered length responses, we need to recover full reference allele from follow up query
+        // follow up query gives correct reference allele for referenced genome positions
+        if (responseReferenceAllele.length() != providedReferenceAllele.length()) {
+            // for altered length deletions, e.g. '1,123,124,TC,T': responseReferenceAllele='C', providedReferenceAllele='TC', length not equal
+            // for altered length del-ins, e.g. '1,123,124,TC,TAA': responseReferenceAllele='C', providedReferenceAllele='TC', length not equal
+            // for altered length insertion but > 1 matching alleles, e.g. '1,123,123,TC,TCA': responseReferenceAllele='-', providedReferenceAllele='TC', length not equal
+            return true;
+        }
+        else if(responseReferenceAllele.length() == providedReferenceAllele.length() && responseReferenceAllele.equals("-")) {
+            // for altered length insertion with one matching allele, e.g. '1,123,123,T,TC': responseReferenceAllele='-', providedReferenceAllele='C'
```
zhx828, line 175:
```suggestion
            // for altered length insertion with one matching allele, e.g. '1,123,123,T,TC': responseReferenceAllele='-', providedReferenceAllele='T'
```

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedGenomicLocationAnnotationServiceImpl.java
```diff
@@ -161,14 +161,31 @@ private VariantAnnotation verifyOrFailAnnotation(VariantAnnotation annotation)
...
+            // ReferenceAllele lengths are equal but also need followup query, so need to check if responseReferenceAllele is '-'
+            return true;
+        }
```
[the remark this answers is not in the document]
zhx828: I think it's fine to separate. Adding comments to differentiate scenarios makes sense to me.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedGenomicLocationAnnotationServiceImpl.java
zhx828: @sheridancbio thoughts?

zhx828 (approved):
Looks good. Just few minor comments.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedGenomicLocationAnnotationServiceImpl.java
```diff
@@ -161,14 +161,31 @@ private VariantAnnotation verifyOrFailAnnotation(VariantAnnotation annotation)
...
+    {
+        // for altered length responses, we need to recover full reference allele from follow up query
+        // follow up query gives correct reference allele for referenced genome positions
+        if (responseReferenceAllele.length() != providedReferenceAllele.length()) {
+            // for altered length deletions, e.g. '1,123,124,TC,T': responseReferenceAllele='C', providedReferenceAllele='TC', length not equal
+            // for altered length del-ins, e.g. '1,123,124,TC,TAA': responseReferenceAllele='C', providedReferenceAllele='TC', length not equal
+            // for altered length insertion but > 1 matching alleles, e.g. '1,123,123,TC,TCA': responseReferenceAllele='-', providedReferenceAllele='TC', length not equal
+            return true;
+        }
+        else if(responseReferenceAllele.length() == providedReferenceAllele.length() && responseReferenceAllele.equals("-")) {
```
sheridancbio, line 174: The conditional checks here seem redundant. I think the code is equivalent to:
```
if (not X) {
    ...;
} else if (X) {
    if (Y) {
        ...;
    }
}
```

Also, the function exits at the first return statement, so we can simplify the logic I think:
```
if (lengths_not_equal) {
    return true;
}
if (responseReferenceAllele.equals("-") {
    return true;
}
return false;
```
Similar to not having an else for the final "return false", we don't need an else for the second test and we know that the lengths are equal (or we would have exited during the first test).

Also please add a space after the keyword "if".

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedGenomicLocationAnnotationServiceImpl.java
sheridancbio: The main purpose of this verification is to make sure that if the user query is specifying reference allele nucleotides that they match the nucleotides in the reference genome. The real concern here is about deletions ... so if the user query says "deleted nucleotides ACC at positions 34782938-34782940" we wanted to not give annotation based on removing (from the reference genome which was different from what the user was basing their variant calls on) nucleotides TAG present at those positions in the VEP database reference genome. There were some cases encountered like this in annotations for AACR Genie samples I believe. Events represented as "insertion-deletions" also involve deletions so they were covered here as well.

Pure insertions basically specify no reference genome nucleotides (in our MAF format at least), so they are not of concern here. SNPs do specify a reference genome nucleotide ... so it would make sense to expand this logic (of checking the reference allele against the VEP database reference genome) I think. When we receive the response from VEP for SNPs I think it includes both the reference allele and the variant allele in the response so we would not need to make a follow up query .. we could instead compare the reference genome nucleotide to the nucleotide specified in the user query. I think adding that check would be good. It may be the case that when we convert MAF into genome region format for SNP variants we lose the reference allele details. I think I remember that some simplification in the formats involved dropping the reference allele ... so maybe that is why we cannot test it here (because the input to genome nexus has dropped the SNP reference nucleotide to conform to the updated format).

sheridancbio (approved):
Looks good. Nice work covering cases of insertion-deletion which were not covered and adding the unit test cases.

It seems that the return reference allele of "-" for insertions was being compared to the (longer) reference allele submitted with a shared prefix (TC -> TCA). These use cases are not encountered during our automatic annotation pipeline runs because we strip shared prefixes before annotation. But this PR now covers them.

Small suggestion and whitespace request below.

PR #677, opened 2023-04-11
Create Enum Value for `fields` Parameter of `/annotation` Endpoint, Improve Swagger Doc

jeffquinn-msk:
- Create enum value `AnnotationType` and use it in place of the `String` type
- Refactor all function definitions
- Improve api docstrings by showing all valid values in the `fields` parameter
- Remove `-XX:MaxPermGen` argument from surefire configuration (there is no PermGen memory in the JVM anymore.. has been that way for a long time, I assume this is old copy pasta and on newer JVMs it actually prevents tests from running, for JVM 8 it gives a warning that this argument has no effect)

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/AnnotationType.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/AnnotationType.java
@@ -1,0 +1,45 @@
+package org.cbioportal.genome_nexus.model;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+
+public enum AnnotationType {
+    ANNOTATION_SUMMARY("annotation_summary"),
+    CLINVAR("clinvar"),
+    HOTSPOTS("hotspots"),
+    MUTATION_ASSESSOR("mutation_assessor"),
+    MY_VARIANT_INFO("my_variant_info"),
+    NUCLEOTIDE_CONTEXT("nucleotide_context"),
+    ONCOKB("oncokb"),
+    PTMS("ptms"),
+    SIGNAL("signal");
+
+    private final String value;
+
+    AnnotationType(String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    @JsonCreator
+    public static AnnotationType fromString(String value) {
+        for (AnnotationType field : AnnotationType.values()) {
+            if (field.value.equalsIgnoreCase(value)) {
+                return field;
+            }
+        }
+        throw new IllegalArgumentException("Invalid value for AnnotationType: " + value);
+    }
+
+    public static String getHumanReadableValues() {
+        return Arrays.stream(AnnotationType.values())
+            .map(AnnotationType::getValue)
+            .collect(Collectors.joining(", "));
+    }
+}
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/AnnotationType.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/AnnotationType.java
@@ -36,10 +36,4 @@ public enum AnnotationType {
         }
         throw new IllegalArgumentException("Invalid value for AnnotationType: " + value);
     }
-
-    public static String getHumanReadableValues() {
-        return Arrays.stream(AnnotationType.values())
-            .map(AnnotationType::getValue)
-            .collect(Collectors.joining(", "));
-    }
 }
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/AnnotationType.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/AnnotationType.java
@@ -2,9 +2,6 @@ package org.cbioportal.genome_nexus.model;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 
-import java.util.Arrays;
-import java.util.stream.Collectors;
-
 
 public enum AnnotationType {
     ANNOTATION_SUMMARY("annotation_summary"),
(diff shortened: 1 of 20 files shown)


model/src/main/java/org/cbioportal/genome_nexus/model/AnnotationType.java
```diff
@@ -0,0 +1,36 @@
+package org.cbioportal.genome_nexus.model;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+
+public enum AnnotationType {
```
onursumer, line 6: Not sure if `Type` is the right name for this enum. `Field` may not be the best name either but maybe for now keep it as is?

```suggestion
public enum AnnotationField {
```

service/src/main/java/org/cbioportal/genome_nexus/service/SelectedAnnotationService.java
```diff
@@ -42,15 +43,15 @@
...
 
     public VariantAnnotation getAnnotation(
-            String variant,
-            String isoformOverrideSource,
-            Map<String, String> token,
-            List<String> fields) throws VariantAnnotationNotFoundException, VariantAnnotationWebServiceException;
+        String variant,
+        String isoformOverrideSource,
+        Map<String, String> token,
+        List<AnnotationType> annotationTypeList) throws VariantAnnotationNotFoundException, VariantAnnotationWebServiceException;
```
onursumer, line 49: I think we should keep the param name as is for consistency

```suggestion
        List<AnnotationType> fields) throws VariantAnnotationNotFoundException, VariantAnnotationWebServiceException;
```

service/src/main/java/org/cbioportal/genome_nexus/service/VariantAnnotationService.java
```diff
@@ -47,7 +47,7 @@
     VariantAnnotation getAnnotation(String variant)
         throws VariantAnnotationNotFoundException, VariantAnnotationWebServiceException;
     List<VariantAnnotation> getAnnotations(List<String> variants);
-    VariantAnnotation getAnnotation(String variant, String isoformOverrideSource, Map<String, String> token, List<String> fields)
+    VariantAnnotation getAnnotation(String variant, String isoformOverrideSource, Map<String, String> token, List<AnnotationType> annotationTypeList)
```
onursumer, line 50: same as above, and for the rest of the PR where applicable. Let's keep the param as is for now?

```suggestion
    VariantAnnotation getAnnotation(String variant, String isoformOverrideSource, Map<String, String> token, List<AnnotationType> fields)
```

onursumer:
@jeffquinn-msk I was wrong. Seems like tests are actually passing for other PRs currently. See, for example, #681. I will investigate further why this one is consistently failing.

onursumer:
@jeffquinn-msk I think I figured out why tests are failing. This PR changes the way the endpoints handles the `fields` param. Somehow it cannot handle lowercase values anymore.

annotation_fields

The tests are still using the `values` of the enums. For example, see https://github.com/jeffquinn-msk/genome-nexus/blob/2d9aacf036e1ba66bdc483b02959dedd29fde2db/web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java#L39. That is `annotation_summary` instead of `ANNOTATION_SUMMARY` for this example.

We should fix this by allowing both lowercase and uppercase values before merging this PR. That is the current behavior in production anyway.

web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
```diff
@@ -35,8 +36,9 @@
 
     private final static String BASE_URL_GENOMIC_LOCATION = "http://localhost:38896/annotation/genomic/";
 
-    private final static String ALL_ENSEMBL_FIELDS = "fields=annotation_summary,nucleotide_context";
-    private final static String MY_VARIANT_INFO_FIELD = "fields=my_variant_info";
+    private final static String ALL_ENSEMBL_FIELDS = "fields=" + AnnotationField.ANNOTATION_SUMMARY.getValue()
+        + "," + AnnotationField.NUCLEOTIDE_CONTEXT.getValue();
+    private final static String MY_VARIANT_INFO_FIELD = "fields=" + AnnotationField.MY_VARIANT_INFO.getValue();
```
onursumer, line 41: I think it is better if we do not modify these string values for integration test. Let's keep it as is.

service/src/test/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationServiceTest.java
```diff
@@ -159,8 +158,8 @@ public void getMutationAssessorEnrichedAnnotationByVariantString()
         this.mockMutationAssessorServiceMethods(variantMockData, maMockData);
         this.mockEnsemblServiceMethods();
 
-        List<String> fields = new ArrayList<>(1);
-        fields.add("mutation_assessor");
+        List<AnnotationField> fields = new ArrayList<>(1);
+        fields.add(AnnotationField.MUTATION_ASSESSOR);
```
onursumer, line 162: `AnnotationField.MUTATION_ASSESSOR` actually returns an all uppercase string value  `MUTATION_ASSESSOR`. Let's make sure that we can still handle lower case values (e.g: `mutation_assessor`) for those fields.

onursumer:
@jeffquinn-msk Do you think you'll have time to improve this PR so that we still allow case insensitive input for `fields`?

web/src/test/java/org/cbioportal/genome_nexus/web/AnnotationIntegrationTest.java
onursumer: We don't need to keep these as string constants, but keeping them as constants ensures that we don't accidentally break backward compatibility. I think it's a good thing to see a failing test after we make a potentially breaking change. If the change is intended we can then manually update the tests.

onursumer (approved):
Looks good, thank you! How did you fix the case sensitivity issue? Is it handled by `AnnotationFieldEnumConverter`?

PR #681, opened 2023-04-21
Fix converting dels using region converter

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/143

### Issue:
When GN is setup using a local VEP, the variants need to be converted to VEP's region format. 
GN converts `11:g.2133018del` to genomic location `11,2133018,2133018,-,-` Then the genomic location is converted to region format `11:2133019_2133018:1/-`, which is wrong.

The reason is genomic location with ref allele "-" are treated as an ins

### Changes:
- Check that the variant allele is "-" before checking that reference allele is "-". In the original code, the HGVSg del was treated as an insertion because the ref allele was "-".
- The package definition for `SelectedAnnotationServiceImpl.java` was fixed

### DIFF
--- a/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java
+++ b/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java
@@ -221,6 +221,18 @@ public class NotationConverter {
         // Ensembl uses a one-based coordinate system
         if (start < 1) {
             region = null;
+        // If the var allele is "-", then it is a deletion
+        // A ref allele of "-" could be an insertion or deletion, so this check should come first.
+        } else if (var.equals("-") || var.length() == 0) {
+            /*
+            Process Deletion
+            Example deletion: 1 206811015 206811016  AC -
+            Example output:   1:206811015-206811016:1/-
+
+            Example deletion: 11 2133018 2133018 - -
+            Example output: 11:2133018-2133018:1/-
+            */
+            region = chr + ":" + start + "-" + end + ":1/-";
         } else if (ref.equals("-") || ref.length() == 0 || ref.equals("NA") || ref.contains("--")) {
             /*
             Process Insertion
@@ -233,13 +245,6 @@ public class NotationConverter {
             catch (NumberFormatException e) {
                 return null;
             }
-        } else if (var.equals("-") || var.length() == 0) {
-            /*
-            Process Deletion
-            Example deletion: 1 206811015 206811016  AC -
-            Example output:   1:206811015-206811016:1/-
-            */
-            region = chr + ":" + start + "-" + end + ":1/-";
         } else if (ref.length() > 1 || var.length() > 1) {
             /*
             Process ONP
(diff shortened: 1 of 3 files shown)


calvinlu3:
@sheridancbio Why did the team choose to use the forward strand, which required the insertion's start,end positions to be flipped? Could we use the reverse strand?

leexgh (approved):
LGTM, thanks!

sheridancbio:
> > @sheridancbio Why did the team choose to use the forward strand, which required the insertion's start,end positions to be flipped? Could we use the reverse strand?
> 
> @calvinlu3 could you clarify what you mean there? We don't really incorporate strandedness in Genome Nexus, so we always use forward strand

I'm not sure I have a definite answer @calvinlu3 ... the genome nexus interface was designed before I joined the project. 

The main use case which motivated the construction of genome nexus was the annotation of MAF files for cBioPortal studies. And in the MAF file (which contains variants which need to be annotated), the gene involved has already been identified (either by hugo symbol or entrez gene id) previous to the annotation step. I believe that there are cases of genes which overlap each other, one on the positive strand and one on the negative strand. So I believe it is true that if you only provide locations without strands it is possible that it may be ambiguous which gene you are asking for variant effect predictions for (either the one on the positive strand or the one on the negative strand). However, I do know that we successfully retrieve variant effects on queries for genes which live on both positive and negative strands even though we are only providing non-stranded locations without strand in our queries. So what I suspect is happening is that the VEP program (running as part of the Genome Nexus processing) returns all the protein alteration effects from both strands ... leaving it up to the client to interpret or select which effects are relevant to them.

So I guess the bottom line is that it seems like the strand is not needed during the query/request because the response contains variant effect on genes on either strand anyway. In our use case we know which gene was specified in the MAF file, so we can select from the response the canonical transcript for that gene (which determines which strand we are paying attention to). I would need to do a little testing to verify that my picture here is accurate. I wonder if this sounds right to you ... is there anything off-track in my thinking?

sheridancbio:
@calvinlu3 Yes, I can confirm it. We followed the REST-style regions documentation here:
https://useast.ensembl.org/info/docs/tools/vep/vep_formats.html#region
which references:
https://useast.ensembl.org/info/docs/tools/vep/vep_formats.html#default
This specifies that "An insertion (of any size) is indicated by start coordinate = end coordinate + 1"
There were errors occurring prior to incorporating this logic, so we went back to the spec and obeyed this rule.

calvinlu3:
> @calvinlu3 Yes, I can confirm it. We followed the REST-style regions documentation here: https://useast.ensembl.org/info/docs/tools/vep/vep_formats.html#region which references: https://useast.ensembl.org/info/docs/tools/vep/vep_formats.html#default This specifies that "An insertion (of any size) is indicated by start coordinate = end coordinate + 1" There were errors occurring prior to incorporating this logic, so we went back to the spec and obeyed this rule.

Thanks, that makes sense. I added a [comment](https://github.com/genome-nexus/genome-nexus/blob/5c4d99dd82ec7865ca1b870dc911bd88fd38f61b/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java#L243-L244) based on what we discussed.

Issue #678, opened 2023-04-13
OncoKB Annotations Are Not Produced


jeffquinn-msk:
I've noticed that the API currently never returns any OncoKB annotation in its resopnse document, even when I submit a variant explicitly from the oncoKB database and make sure to include my oncokb api key in the request, which I have confirmed independently is valid and working.

Even running the OncoKB test curl command (https://github.com/genome-nexus/genome-nexus/blob/master/.github/workflows/oncokb_test.yml ) with my own OncoKB key I'm not able to get any OncoKB annotation in the response document. However I am confused because according to github actions history, this test has always been passing and is still currently passing.

My hypothesis is that the Github action currently only succeeds because of a flaw in the shell script if statement:

```
if [ `grep -q "$SEARCH_HTTP_CODE" "$FILE"` -a `grep -q "$SEARCH_ONCOKB" "$FILE"` ]
then
  echo "Request succeeded"
  exit 0
else
  echo "Request failed"
  exit 1
fi
```

I'm not sure you can use this combination of POSIX test (`[`) and backticks, it seems like its not doing what the author intended, and will always execute the echo "success" block. To wit, I've conducted this experiment (on both sh and bash):

```
sh-3.2$ if [ `true` -a `true` ]; then echo "success"; else echo "fail"; fi
success
sh-3.2$ if [ `true` -a `false` ]; then echo "success"; else echo "fail"; fi
success
sh-3.2$ if [ `false` -a `false` ]; then echo "success"; else echo "fail"; fi
success
sh-3.2$ if [ `false` -a `true` ]; then echo "success"; else echo "fail"; fi
success
```
(`true` is a builtin shell command that always returns successfully, `false` always fails with status)

POSIX test (`[`) is super annoying and prone to bugs so I'd recommend we rewrite this status check to use the bash conditional builtin (`[[`).

Anyway, this explains why the behavior I see is not consistent with the passing oncokb test.

To further debug I have run genome-nexus locally following the readme instructions. When I hit the annotation endpoint I see that oncokb is failing at this point:

https://github.com/genome-nexus/genome-nexus/blob/df3857eec809b7db72423ae36d8d5079bc7f96fe/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/OncokbResolver.java#L21

This is because `transcriptConsequenceSummary.getEntrezGeneId()` returns null.

[...]

PR #687, opened 2023-06-13
Update oncokb java api client and mapper configs

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/678

- We need to update oncokb java api client to match their response. 
- Updated oncokb test.

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/transformer/ExternalResourceTransformer.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/transformer/ExternalResourceTransformer.java
@@ -1,6 +1,9 @@
 package org.cbioportal.genome_nexus.service.transformer;
 
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.json.JsonMapper;
 import com.mongodb.DBObject;
 import org.cbioportal.genome_nexus.service.ResourceTransformer;
 import org.cbioportal.genome_nexus.service.exception.ResourceMappingException;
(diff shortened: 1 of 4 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/transformer/ExternalResourceTransformer.java
```diff
@@ -50,11 +53,8 @@ private List<T> mapJsonToInstance(DBObject rawJson, Class<T> type, ObjectMapper
     {
         List<T> list = new ArrayList<>();
         ObjectMapper mapper = objectMapper;
-
-        if (mapper == null)
-        {
-            mapper = new ObjectMapper();
-        }
+        mapper.enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS);
```
leexgh, line 56: I need to add this because there is an enum mapping issue:
`"Cannot deserialize value of type `org.oncokb.client.Query$ReferenceGenomeEnum` from String "GRCh37": not one of the values accepted for Enum class: [GRCH37, GRCH38] at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.oncokb.client.IndicatorQueryResp["query"]>org.oncokb.client.Query["referenceGenome"])"`

leexgh:
@jeffquinn-msk Yes the problem is in deserializing. I believe the failing test is not related, it failed in other pull requests before.

qwwqwwq:
The last PR made that was merged, which was exclusively a change to markdown documentation files, had two builds run on circleCI, one where this same test case failed and one where it passed: https://app.circleci.com/pipelines/github/genome-nexus/genome-nexus/87

That indicates to me that the test is useless, or there is something very very wrong with the surefire configuration:

Taking a quick look at it:

```
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-surefire-plugin</artifactId>
        <version>3.0.0-M3</version>
        <configuration>
          <forkCount>3</forkCount>
          <reuseForks>true</reuseForks>
          <argLine>-Xmx1024m</argLine>
          <systemPropertyVariables>
            <projectVersion>1.0</projectVersion>
          </systemPropertyVariables>
        </configuration>
      </plugin>
```

it seems a bit suspicious to me, running integration tests in parallel is a notorious source of flakiness, even more so when they involves a real database (as these tests do).

All that being said I have no idea how this particular error could result from a race condition since it seems like the test should never pass given in the input.

Anyway tl;dr I can't approve a PR when the tests are failing, but happy to be overruled if you really aren't concerned. One option to proceed that would get my approval would be to slap the old `@Ignore` on this test case for now.

leexgh:
I have another [PR](https://github.com/genome-nexus/genome-nexus/pull/691) to fix the tests, merge this one

PR #689, opened 2023-06-28
Read reVUE json from reVUE-data repository

leexgh:
Changes in this PR:
- New config property - `revue.url`: Read reVUE from https://github.com/knowledgesystems/reVUE-data/blob/main/VUEs.json. Instead of reading from local json file, we could set `revue.url` to point to a specific data file. So we don't need to make new images for every reVUE data updates.
- New config property - `overwrite_by_confirmed_revue_only`: Default value is true, if `overwrite_by_confirmed_revue_only = true`, overwrite only confirmed VUEs(vue). If `overwrite_by_confirmed_revue_only = false`, overwrite by both confirmed and unconfirmed VUEs.
- Remove local json reading code
- Read and parse json from `revue.url`

Test: http://localhost:38080/annotation/5:g.112151184A>G?fields=annotation_summary
The response should contain reVUE
```
"vues": {
      "hugoGeneSymbol": "APC",
      "genomicLocationDescription": "5:g.112151184A>G",
      "defaultEffect": "synonymous",
      "comment": "Introduces splice acceptor site causing a frameshift",
      "pubmedId": 29316426,
      "referenceText": "Yaeger et al., 2018",
      "variant": "5:g.112151184A>G",
      "genomicLocation": "5,112151184,112151184,A,G",
      "transcriptId": "ENST00000257430",
      "revisedProteinEffect": "p.G279Ffs*10",
      "revisedVariantClassification": "Frame_Shift_Ins",
      "context": "Recurrent alteration in CRC",
      "vepPredictedProteinEffect": "p.*279*",
      "vepPredictedVariantClassification": "Splice_Region",
      "confirmed": true
    }
```


leexgh:
The test `MyVariantInfoIntegrationTest.testMyVariantInfoAnnotationLargePOST` has been failed for a while, I'll fix it in a separate PR.

service/src/main/java/org/cbioportal/genome_nexus/service/remote/RevueDataFetcher.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.RestTemplate;
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
+
+@Component
+public class RevueDataFetcher {
+
+    private static final String revueUrl = "https://raw.githubusercontent.com/knowledgesystems/reVUE-data/main/VUEs.json";
```
inodb, line 12: @leexgh should this be some configurable property? E.g. if we want revue-beta.genomenexus.org to point to a dev version of the reVUE JSON? And maybe we could have some property whether or not to include validated VUEs only or all VUEs?

leexgh: Yes, I added two properties: `revue.url` and `overwrite_by_confirmed_revue_only`

inodb (approved):
Looks great- thanks so much!

PR #693, opened 2023-07-17
Update VUEs.json to be in sync with new reVUE-data repo file

leexgh:
This is a temporary update, will use reVUE-data repo file directly in the following PR


leexgh:
Decide to close this one and go with reading from reVUE-data PR directly

PR #701, opened 2023-08-15
Harmonize genomic location coordinates

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/702
This PR can also fix some failing variants from: https://github.com/genome-nexus/annotation-tools/issues/57 
And some failing variants on GENIE genome nexus: https://github.com/genome-nexus/genome-nexus-annotation-pipeline/issues/260

Harmonize end position for some of the variants that have
- no end position
- end position < start position
- start - end does not equal to length of reference allele

Todo:
- [x] Harmonize genomic location coordinates
- [x] Return a brief reason of genomic location harmonization

### DIFF
--- a/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java
+++ b/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/NotationConverter.java
@@ -1,7 +1,5 @@
 package org.cbioportal.genome_nexus.component.annotation;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 import org.cbioportal.genome_nexus.model.GenomicLocation;
 import org.cbioportal.genome_nexus.util.GenomicVariant;
 import org.cbioportal.genome_nexus.util.GenomicVariantUtil;
@@ -18,7 +16,6 @@ import java.util.Map;
 
 @Component
 public class NotationConverter {
-    private static final Log LOG = LogFactory.getLog(NotationConverter.class);
 
     public static final String DEFAULT_DELIMITER = ",";
 
@@ -133,9 +130,8 @@ public class NotationConverter {
                 nStart -= 1;
             }
             start = nStart;
-            LOG.info("Start position is changed from " + genomicLocation.getStart() + " to " + start + " for genomic location: " + genomicLocation.getChromosome() + "," + genomicLocation.getStart() + "," + genomicLocation.getEnd() + "," + genomicLocation.getReferenceAllele().trim() + "," + genomicLocation.getVariantAllele().trim() + ". Reason: remove common prefix alleles.");
         }
-        end = harmonizeGenomicLocationCoordinate(genomicLocation, chr, start, end, ref, var);
+        end = harmonizeGenomicLocationCoordinate(start, end, ref);
         normalizedGenomicLocation.setStart(start);
         normalizedGenomicLocation.setEnd(end);
         normalizedGenomicLocation.setReferenceAllele(ref);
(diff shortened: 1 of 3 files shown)


inodb (approved):
LGTM - thank you!!

sheridancbio:
The solution proposed is to ignore / overwrite the genomic coordinate for endPosition in the MAF file based on the Reference Allele string length.

In overwriting the endPosition, the startPosition is assumed to be accurate. If we are assuming that the client/user/variant-caller has supplied an erroneous endPosition, I wonder how we can assume that they are providing an accurate start position. Similarly, they have provided a Reference Allele String which is discordant with the coordinates --- when there is a discrepancy between the start/end coordinates and the Reference Allele String, how do we know that it is not the Reference Allele which is incorrect ?

I am only pointing out potential problems with making these conversions. The HGVS representation "throws away" the reference allele string under the assumption that the coordinates will define the reference allele and so the information in the nucleotide sequence is redundant. But here our input data is self-contradicting. I am not sure we can be certain that we will not be altering the true variation found by the sequencer / variant calling algorithm.

To make it concrete with an example, imagine input:
chr: 16
startPos: 3843465
endPos: 3843466
RefAllele: A
AltAllele: ACGTA

Here the start/end positions indicate two nucleotides in the genome but the RefAllele indicates just one. The hg19 reference genome in this area can be seen here:
https://genome.ucsc.edu/cgi-bin/hgTracks?db=hg19&lastVirtModeType=default&lastVirtModeExtraState=&virtModeType=default&virtMode=0&nonVirtPosition=&position=chr16%3A3843462-3843471

leading up to the variant the reference genome has nucleotides AGC in positions 2843462, 2843463, 2843464
position 3843465 has nucleotide A in the genome
position 3843466 has nucleotide G in the genome
after that the following genomic alleles are GCCCG

If we take the start and end positions from the input as being correct then we can interpret the variant caller as having sequenced something like:
AGC **ACGTA** GCCCG
which corresponds to a deletion of the AG at the two positions specified and an insertion of "ACGTA" in place of the deleted segment.

However if we take the reference sequence as being accurate, then we assume that only the "A" at position 3843465 was deleted and the AltAllele sequence was inserted there ... so the sequencing would have looked like:
AGC **ACGTA** GGCCCG

And the "G" nucleotide in the reference genome at position 3843466 is still considered to be present.

But without knowing what the sequencer detected I think we can have some doubts about what actually was detected. Which of the following did the sequencer see?
AGCACGTAGCCCG
AGCACGTAGGCCCG

A net gain of 3 nucleotides likely has a different effect than a net gain of 4.

Maybe I am overgeneralizing with this example? It seems to me that a general rule of "take the length of the reference allele as more authoritative than the start/end coordinate span" is not safe unless we know that we can always trust the reference allele string as being the correct definition of what in the reference genome was altered.

I just wanted to raise this concern for your consideration.

leexgh:
@sheridancbio Thanks for looking into it! 

@inodb and I had a discussion and we believe it still makes sense to do it, here are my two cents:
- This is a fix requested by GENIE, and they need it by this Friday 09/01 to make a new release. The data comes from many institutes and sometimes it's a bit hard to let the providers correct the input data.
- It only fixes some corner cases, like https://github.com/genome-nexus/annotation-tools/issues/55 (in this ticket it's because end < start, which is clearly wrong), or some rare cases in which the end position is missing. For most of the queries it doesn't change anything. It's not the best solution, but still good to have to at least fix some of the failing cases. 
- There is an optional parameter in annotation pipeline `--add-original-genomic-location`, this will add original genomic location data columns into the output, name columns with prefix `'IGNORE_Genome_Nexus_Original_'`. In this way people can still have a copy of their original input.
- I will work with @rmadupuri  to make sure future maf annotations do that too.

I'd also do these followup tasks to improve:
- Add a message in the response to tell people whether the genomic location is changed, how, what, and why it is changed. (e.g. `"genomicLocationExplanation":"End position changes from 123 to 122, because end position should be the position of last deleted nucleotide."`). This information will also be added into the annotation pipeline as an optional column. People could do for instance `-m` to see this message in maf for variants that have genomic location changed. So people can see what was done. If they investigate the failing variants, the message column would be a good way to provide our "feedback" to user as well, so they could know if it's possible to fix the data on their end upstream to reduce failing.
- Improve error message in the response. I have a PR: https://github.com/genome-nexus/genome-nexus/pull/704 still in progress (e.g. `"errorMessage": "Unable to parse HGVS notation '12:g.58142319A>C': Reference allele extracted from 12:58142319-58142319 (T) does not match reference allele given by HGVS notation 12:g.58142319A>C (A)"`
). I'll add other error messages as well like failing on annotation verification (cases like provided ref doesn't equal to annotated ref). This will also be added to annotation pipeline as an optional column along with the "genomicLocationExplanation", so people can know more about the failure reason and if that's because of the input data.

sheridancbio:
Ok @leexgh - that sounds reasonable. As long as we communicate the alterations which were performed in making a variant effect prediction, then the caller / user will know that the predicted effect is based on a query string which was modified. In this case, I would also recommend communicating back to genie the alterations which occurred in these corner cases and suggest to them that the source institutions which did the sequencing check these cases and verify the alteration was proper based on the sequencing runs. (unless we can establish that all corner cases were based on a processing artifact affecting the endPosition which is uncovered and we have confidence that the startPosition and reference allele string as supplied can be trusted as authoritative).

PR #710, opened 2023-09-19
Support missing end position

leexgh:
This is a follow up PR of https://github.com/genome-nexus/genome-nexus/pull/701. There is an issue in post request with missing end position, this PR fixes the issue and supports missing end position in post and get


inodb (commented):
LGTM - thank you!

inodb (approved):

Issue #448, opened 2020-09-18
Allele "N" is not being annotated


yichaoS:
VEP is able to annotate allele 'N'. Seems like GN/annotator is not accepting it - no annotation returned.							

Example: 
in ov_pancan
No result when querying GN: 
http://annotation.genomenexus.org/annotation/genomic/1%2C205068920%2C205068922%2CGAT%2CANA?isoformOverrideSource=uniprot&fields=hotspots%2Cmutation_assessor

VEP returned result: http://grch37.ensembl.org/Homo_sapiens/Tools/VEP/Results?tl=9cdceSKfKPbBAc66-6575311
<img width="871" alt="Screen Shot 2020-09-18 at 5 00 00 PM" src="https://user-images.githubusercontent.com/5973438/93644592-5141c500-f9d0-11ea-81ef-415d18518653.png">

For more examples and complete "mutated" variants stats, please check here:
https://docs.google.com/spreadsheets/d/1RFj8zzMPv4PCAs3sFM9ro3FP4vRg-Za-JI2AFlx6FHg/edit#gid=1944280727

sbabyanusha:
Updated VEP return Link: https://grch37.ensembl.org/Homo_sapiens/Tools/VEP/Results?tl=jHzc3RYzAUWhB27T-9525994

No result when querying GN: https://www.genomenexus.org/annotation/genomic/1%2C205068920%2C205068922%2CGAT%2CANA?fields=annotation_summary

inodb:
If we fix the annotator to provide variant classification most of these mutations would be filtered out (intronic etc)

leexgh:
Removing `!id.contains("-")` from https://github.com/genome-nexus/genome-nexus/blob/master/service/src/main/java/org/cbioportal/genome_nexus/service/cached/CachedVariantAnnotationFetcher.java#L37C16-L37C33 can fix the failing annotation

PR #714, opened 2023-10-06
Support N allele

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/448


leexgh:
@averyniceday @sheridancbio Do you think removing the "N" check is ok? Not sure if you ever see any "N" alleles

Issue #725, opened 2023-12-13
Annotation pipeline succeeds but returns empty results for intergenic variants


rmadupuri:
Example, 1:g.69918522A>T is an intergenic variant and the annotation pipeline returns empty. When importing to the portal, the variants are filtered out by the Variant_Classification column. Any variant that falls under the "Silent", "Intron", "3'UTR", "5'UTR", "3'Flank", "5'Flank", "IGR" category is filtered out and the intergenic_variants get imported as MUTATED into the Database. 

Although there is no transcript info https://www.genomenexus.org/annotation/1:g.69918522A%3ET?fields=hotspots,annotation_summary,my_variant_info,clinvar,signal,mutation_assessor,mutation_assessor,mutation_assessor,mutation_assessor it would be helpful to use the consequence term and fill the Variant_Classification so the intergenic variants gets filtered out on import.

leexgh:
Solution: put `IGR` to `Variant_Classification` and `intergenic_variant` to `Consequence`

PR #746, opened 2024-07-16
Add intergenicConsequnceSummaries and variant classification in annotationSummary

leexgh:
Fix:https://github.com/genome-nexus/genome-nexus/issues/725
**Background**: Intergenic_variant has no transcript so it returns empty in genome-nexus-annotation-pipeline annotation, and they are not supposed to be imported into cbioportal. 
The variant classification is resolved from transcript consequence terms. So when variants don't have transcripts, the variant classification is empty. And if variant classification is empty, the variant gets imported into cbioportal because we filter variants by variant classification == IGR / etc...
- [x] Intergenic_variant should be filtered out when importing into cbioportal, so we need to set 'IGR' variant classification for those variants.
    - [x] Add a dew field `intergenicConsequenceSummaries` in annotationSummary
    - [x] Add `variantClassification` in `intergenicConsequenceSummary`

- [x] Update transcript consequence list (https://useast.ensembl.org/info/genome/variation/prediction/predicted_data.html) and their variant classification

Example 1:g.69918522A>T:
Previous:
```
"annotation_summary": {
    "variant": "1:g.69918522A>T",
    "genomicLocation": {
      "chromosome": "1",
      "start": 69918522,
      "end": 69918522,
      "referenceAllele": "A",
      "variantAllele": "T"
    },
    "strandSign": "+",
    "variantType": "SNP",
    "assemblyName": "GRCh37",
    "transcriptConsequences": [],
    "transcriptConsequenceSummaries": []
  }
```
New
```
"annotation_summary": {
    "variant": "1:g.69918522A>T",
    "genomicLocation": {
      "chromosome": "1",
      "start": 69918522,
      "end": 69918522,
      "referenceAllele": "A",
      "variantAllele": "T"
    },
    "strandSign": "+",
    "variantType": "SNP",
    "assemblyName": "GRCh37",
    "transcriptConsequences": [],
    "transcriptConsequenceSummaries": [],
    "intergenicConsequenceSummaries": [
      {
        "impact": "MODIFIER",
        "variantAllele": "T",
        "consequenceTerms": [
          "intergenic_variant"
        ],
        "variantClassification": "IGR"
      }
    ]
  }
```

Discussion:
1. Do we want to use the same order of severity from Ensembl? (According Ensembl: This ordering is necessarily subjective and API and VEP users can always get the full set of consequences for each allele and make their own severity judgement. ) For example, in Genome nexus xx is higher than xx, but the Ensembl takes xx lower than xx

We will keep our original order as long as it's in the same impact group (high, moderate, low, modifier)

2. Does manually adding `consequenceTerms` and `variantClassification` make sense?  Because the variant doesn't have transcript, but the two fields are under transcriptConsequenceSummary. Should we directly fix this in genome nexus annotation pipeline instead?

We decide to add intergenic_consequences_summary inside annotation_summary

3. Variant classification (especially "Feature_mutation" and "Sequence_mutation") for missing transcript consequence terms

We decide to use the original terms for variant classification, in the future if we have a better name we can replace it

### DIFF
--- a/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/VariantClassificationResolver.java
+++ b/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/VariantClassificationResolver.java
@@ -11,6 +11,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 @Component
 public class VariantClassificationResolver
@@ -45,6 +46,21 @@ public class VariantClassificationResolver
             variantClassification = this.resolveVariantClassification(
                 this.consequencePrioritizer.pickHighestPriorityConsequence(transcriptConsequence.getConsequenceTerms()),
                     variantType, isInframe);
+        }
+        // if variant has no transcript and has intergenicConsequences instead, try to resolve the variant classification from intergenicConsequences
+        else if (variantAnnotation != null && variantAnnotation.getIntergenicConsequences() != null && !variantAnnotation.getIntergenicConsequences().isEmpty()) {
+            // pick the highest priority consequence from the intergenic consequences
+            // get every getConsequenceTerms() from each intergenic consequence and put in a list
+            variantClassification = this.resolveVariantClassification(
+                this.consequencePrioritizer.pickHighestPriorityConsequence(
+                    variantAnnotation.getIntergenicConsequences()
+                    .stream()
+                    .flatMap(consequence -> consequence.getConsequenceTerms()
+                    .stream())
+                    .collect(Collectors.toList())),
+                    variantType,
+                    isInframe
+            );
         }
         // use the most severe consequence to resolve the variant classification
         else if (variantAnnotation != null)
@@ -186,9 +202,13 @@ public class VariantClassificationResolver
         variantMap.put("missense_variant",              "Missense_Mutation");
         variantMap.put("protein_altering_variant",      "Missense_Mutation"); // Not always correct, resolveVariantClassification handles the exceptions
         variantMap.put("coding_sequence_variant",       "Missense_Mutation"); // Not always correct, resolveVariantClassification handles the exceptions
+        variantMap.put("coding_transcript_variant",       "Missense_Mutation"); // Not always correct, resolveVariantClassification handles the exceptions
         variantMap.put("conservative_missense_variant", "Missense_Mutation");
         variantMap.put("rare_amino_acid_variant",       "Missense_Mutation");
         variantMap.put("transcript_amplification",      "Intron");
+        variantMap.put("feature_elongation",      "Feature_Elongation");
+        variantMap.put("feature_truncation",      "Feature_Elongation");
+        variantMap.put("sequence_variant",      "Sequence_Variant");
         variantMap.put("splice_region_variant",         "Splice_Region");
         variantMap.put("splice_donor_region_variant",         "Splice_Region");
         variantMap.put("splice_polypyrimidine_tract_variant",         "Splice_Region");
@@ -216,6 +236,7 @@ public class VariantClassificationResolver
         variantMap.put("regulatory_region_variant",     "IGR");
         variantMap.put("regulatory_region_amplification",     "IGR");
         variantMap.put("regulatory_region",             "IGR");
+        variantMap.put("regulatory_region_ablation",             "IGR");
         variantMap.put("intergenic_variant",            "IGR");
         variantMap.put("intergenic_region",             "IGR");
         variantMap.put("upstream_gene_variant",         "5'Flank");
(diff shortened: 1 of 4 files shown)


component/src/main/java/org/cbioportal/genome_nexus/component/annotation/VariantClassificationResolver.java
```diff
@@ -186,9 +186,13 @@ private static Map<String, String> initVariantMap()
         variantMap.put("missense_variant",              "Missense_Mutation");
         variantMap.put("protein_altering_variant",      "Missense_Mutation"); // Not always correct, resolveVariantClassification handles the exceptions
         variantMap.put("coding_sequence_variant",       "Missense_Mutation"); // Not always correct, resolveVariantClassification handles the exceptions
+        variantMap.put("coding_transcript_variant",       "Missense_Mutation"); // Not always correct, resolveVariantClassification handles the exceptions
         variantMap.put("conservative_missense_variant", "Missense_Mutation");
         variantMap.put("rare_amino_acid_variant",       "Missense_Mutation");
         variantMap.put("transcript_amplification",      "Intron");
+        variantMap.put("feature_elongation",      "Feature_Mutation");
+        variantMap.put("feature_truncation",      "Feature_Mutation");
+        variantMap.put("sequence_variant",      "Sequence_Mutation");
```
leexgh, line 195: Need advice on what variant classification these should be. I put "Feature_Mutation" and "Sequence_Mutation" for now but these are not official names 🙂

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationSummaryServiceImpl.java
```diff
@@ -109,20 +109,31 @@ public VariantAnnotationSummary getAnnotationSummaryForCanonical(VariantAnnotati
...
-                annotationSummary.setVues(this.vuesMap.get(annotationSummary.getTranscriptConsequenceSummary().getTranscriptId()));
+        if (annotationSummary != null) {
+            if (canonicalTranscript == null) {
+                // Set variantClassification and consequence to "IGR" and "intergenic_variant" when variant is "intergenic_variant".
+                // This is useful for genome nexus annotation pipeline, so when doing curation, those mutations will be filtered out by "IGR" rule, instead of being imported as "MUTATED".
+                if (annotation.getMostSevereConsequence() != null && annotation.getMostSevereConsequence().equals("intergenic_variant")) {
+                    annotationSummary.setTranscriptConsequenceSummary(new TranscriptConsequenceSummary());
+                    annotationSummary.getTranscriptConsequenceSummary().setVariantClassification("IGR");
+                    annotationSummary.getTranscriptConsequenceSummary().setConsequenceTerms("intergenic_variant");
+                }
```
leexgh, line 120: Does manually adding consequenceTerms and variantClassification make sense? Because the variant doesn't have transcript, but the two fields are under transcriptConsequenceSummary. Should we directly fix this in genome nexus annotation pipeline instead? https://github.com/genome-nexus/genome-nexus-annotation-pipeline/pull/282

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/VariantClassificationResolver.java
leexgh: We decide to keep original name as the classification

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/VariantClassificationResolver.java
```diff
@@ -46,6 +47,14 @@ public String resolve(VariantAnnotation variantAnnotation, TranscriptConsequence
                 this.consequencePrioritizer.pickHighestPriorityConsequence(transcriptConsequence.getConsequenceTerms()),
                     variantType, isInframe);
         }
+        // if variant has no transcript and has intergenicConsequences instead, try to resolve the variant classification from intergenicConsequences
+        else if (variantAnnotation != null && variantAnnotation.getIntergenicConsequences() != null && !variantAnnotation.getIntergenicConsequences().isEmpty()) {
+            // pick the highest priority consequence from the intergenic consequences
+            // get every getConsequenceTerms() from each intergenic consequence and put in a list
+            variantClassification = this.resolveVariantClassification(
+                this.consequencePrioritizer.pickHighestPriorityConsequence(variantAnnotation.getIntergenicConsequences().stream().flatMap(consequence -> consequence.getConsequenceTerms().stream()).collect(Collectors.toList())),
+                    variantType, isInframe);
```
onursumer, line 56: It would be easier to read this way, but what we actually need is a prettier equivalent for Java.

```suggestion
            variantClassification = this.resolveVariantClassification(
                this.consequencePrioritizer.pickHighestPriorityConsequence(
                    variantAnnotation
                    .getIntergenicConsequences()
                    .stream()
                    .flatMap(consequence -> consequence.getConsequenceTerms().stream())
                    .collect(Collectors.toList())
                ),
                variantType,
                isInframe
            );
```

onursumer (approved):
LGTM!

PR #752, opened 2024-08-13
Update to Mutation Assessor v4

leexgh:
Part of: https://github.com/genome-nexus/genome-nexus/issues/753
New v4 mutation assessor example 7:g.140453136A>T: 
```
"mutation_assessor": {
    "uniprotId": "P15056",
    "sv": 4,
    "hgvspShort": "p.V600E",
    "functionalImpactScore": 7.110925419276407,
    "functionalImpactPrediction": "high",
    "msa": "P15056.4_full_b0.6_rg0.3",
    "mav": 4
  }
```
- We previously fetch mutation assessor v3 data by hgvsg, new v4 data doesn't have hgvsg information, so we use "uniprotId,hgsvgShort" as the key, this format can also avoid compound index slowness. 
- MutationAssessorWebServiceException is removed, because we no longer call mutation assessor through external API. Related tests are deleted as well.
- Change the mutation assessor model to remove the nested `annotation` (`mutation_assessor{annotation{...}}` changes to `mutation_assessor{...}`).
- Remove mutation assessor mixin
- Update mutation assessor test mock data and tests


model/src/main/java/org/cbioportal/genome_nexus/model/MutationAssessor.java
```diff
@@ -1,261 +1,76 @@
...
+    
+    public void setHgvspShort(String hgvspShort) {
+        this.hgvspShort = hgvspShort;
+    }
+    
+    private Double f_score;
+    
+    public Double getF_score() {
+        return f_score;
+    }
```
onursumer, line 41: Not sure if it is a big deal but previously we did something like this to have nicer Java field names.

```Java
@Field("F_score")
public Double getFunctionalImpactScore() {
    return functionalImpactScore;
}
```

Should we do something similar for the new model?

onursumer (commented):
Looks good! Just a question about variable naming.

model/src/main/java/org/cbioportal/genome_nexus/model/MutationAssessor.java
leexgh: @onursumer Good point! Right now I just use whatever in their original data, I guess we can rename to `functionalImpact` and `functionalImpactScore`

PR #759, opened 2024-11-28
Support dup and inv in local vep GN

leexgh:


zhx828:
> The originalVariantQuery is not the actual "original"
Another note, this should return hgvsg as originalVariantQuery in local vep setting instead of the genomic location.

PR #760, opened 2024-12-12
set max limit on cached variant to allign with mongo max key size

bprize15:


persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/JsonMongoRepositoryImpl.java
```diff
@@ -81,6 +81,10 @@ public void saveDBObject(String collection, String key, DBObject value)
         dbObject.put("_id", key);
 
         // save the object into the correct repository
-        this.mongoTemplate.save(dbObject, collection);
+        // NOTE: Mongo has a limit of 1023 bytes as its max key size. When g. is cached, c. will be cached as a side effect.
+        // Since the c. representation will be longer, we need to choose a "safe" limit
+        if (key.getBytes().length < 900) {
+            this.mongoTemplate.save(dbObject, collection);
+        }
```
calvinlu3, line 88: Why is the c. longer than the g. and how did you choose the number 900 for the maximum length?

bprize15: When converting from g. to c., we must attach a transcript, so the alteration name will become longer. We chose 900 as the max length to leave us a wide buffer while also being large enough to still cache most variants.

inodb (approved):

PR #762, opened 2024-12-17
Add hgvs and genomic change examples documents

leexgh:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,14 +76,18 @@ java -jar web/target/web-*.war
 ```
 
 ## Examples
-|Type| HGVS |Genomic change| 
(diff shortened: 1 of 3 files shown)


inodb (commented):
Thanks Xiang!

Maybe good to add some columns for REST API response, variant page? For genomic change: could clarify that it's a comma separated version of MAF (from TCGA)


In terms of the location, should some of this live on the [docs site](https://docs.genomenexus.org/) too? E.g. could do a new section above MAF and VCF, "annotate a variant". Then link from the README to the docs page?

leexgh:
@inodb Doc updated

PR #763, opened 2024-12-23
Fix missing CDKN2A canonical transcript

leexgh:
We have two canonical transcript for `mskcc` isoform override - ENST00000428597 and ENST00000361570, but because ENST00000428597 is the first transcript and we only choose one transcript per gene, so we store this transcript in the database and missed the ENST00000361570. 
This PR add `ENST00000361570` when pulling `mskcc` isoform override canonical transcript id list, so the matching transcript will be selected as canonical.

Example: 9:g.21993799_21993800del
This variant has ENST00000428597 (CDKN2B-AS1) as canonical transcript previously, now it has ENST00000361570 (CDKN2A) as canonical transcript


inodb (approved):
LGTM - thank you!

PR #772, opened 2025-01-31
always use hgvs service for annotations

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/544


inodb (approved):
Looks great! Thanks for doing this

Would be good to update docs a bit more to mention this change (could also be follow up ticket):

- [ ] Update README.md to indicate region is no longer used and elaborate how one should set up VEP with mysql and leverage it in the Docker Compose setup
- [ ] I liked the diagram you drew on the whiteboard about this change. Maybe worthwhile to have some version of that on docs.genomenexus.org ?

bprize15:
> Looks great! Thanks for doing this
> 
> Would be good to update docs a bit more (could also be follow up ticket):
> 
> * [ ]  Update README.md to indicate region is no longer used and elaborate how one should set up VEP with mysql and leverage it in the Docker Compose setup
> * [ ]  I liked the diagram you drew on the whiteboard about this change. Maybe worthwhile to have some version of that on docs.genomenexus.org ?

Sounds good! I will definitely do that. I think I want to get genome-nexus-vep-2 approved, and then I will update the docs and merge the PR. I actually refactored quite a bit more, so I was hoping you could give the PR another look. Really appreciate all the help!

persistence/src/main/java/org/cbioportal/genome_nexus/persistence/internal/SourceInfoRepository.java
```diff
@@ -30,8 +37,24 @@ public GenomeNexusInfo genomeNexusInfo(@Value("${genomenexus.server.version:NA}"
...
-    public VEPInfo vepInfo(@Value("${gn_vep.server.version:NA}") String serverVersion, @Value("${gn_vep.cache.version:NA}") String cacheVersion, @Value("${gn_vep.region.url:}") String vepRegionURL) {
-        return new VEPInfo(serverVersion, cacheVersion, vepRegionURL);
+    public VEPInfo vepInfo(@Value("${vep.url:}") String vepUrl, @Value("${vep.static:true}") boolean isStatic) throws MalformedURLException {
+        URL url = new URL(vepUrl);
+        RestTemplate restTemplate = new RestTemplate();
+        String serverVersion = NA_STRING;
+        String vepVersion = NA_STRING;
+        String comment = null;
+        try {
+            VepVersionObject vepVersionObject = restTemplate.getForObject(url.getProtocol() + "://" + url.getAuthority() + "/info/software", VepVersionObject.class);
```
leexgh, line 47: Can't find where /info/software is.
Also I'm not sure if this is a good practice to call API in repository layer, maybe good to do it in service?

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedVariantAnnotationService.java
```diff
@@ -33,121 +33,147 @@
...
+    private String constructFollowUpQuery(String originalQuery, VariantType variantType)
     {
-        // create a deletion variant covering the referenced genome positions
-        // this code should only run for delins variants where part of the TumorSeq allele matches the reference genome
-        String followUpQuery = originalQuery.replaceFirst("ins.*|del.*","");
-        if (followUpQuery.length() == originalQuery.length()) {
-            return ""; // unexpectantly called constructFollowUpQuery on non-delins hgvs query -- this annotation will fail
+        String followUpQuery = "";
+        if (variantType == VariantType.HGVS) {
+            followUpQuery = originalQuery.replaceFirst("ins.*|del.*|[A|T|C|G]>[A|T|C|G]","");
```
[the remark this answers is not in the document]
leexgh: Does A|T|C|G]>[A|T|C|G] ever need a follow up query?

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedVariantAnnotationService.java
```diff
@@ -33,121 +33,147 @@
...
+            if (followUpQuery.length() == originalQuery.length()) {
+                return "";
+            }
+            followUpQuery += "del";
+        } else if (variantType == VariantType.GENOMIC_LOCATION) {
+            GenomicLocation genomicLocation = notationConverter.parseGenomicLocation(originalQuery);
+            genomicLocation.setVariantAllele("-");
+            if (genomicLocation.getReferenceAllele().equals( "-")) {
```
leexgh, line 171:
```suggestion
            if (genomicLocation.getReferenceAllele().equals("-")) {
```

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedVariantAnnotationService.java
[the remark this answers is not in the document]
bprize15: It does. Without the ensembl variation database, we need to use a follow up query to get the ref allele at that location.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedVariantAnnotationService.java
```diff
@@ -33,121 +33,147 @@
...
+    private VariantAnnotation verifyOrFailAnnotation(VariantAnnotation annotation, VariantType variantType)
     {
-        String originalVariantQuery = annotation.getOriginalVariantQuery(); // save for failed response
-        String originalVariant = annotation.getVariant(); // save for failed response
-        String originalQuery = originalVariantQuery;
-        if (originalVariantQuery == null || originalVariantQuery.length() == 0) {
-            originalQuery = originalVariant;
+        String ref = "";
+        if (variantType == VariantType.HGVS) {
+            ref = GenomicVariantUtil.providedReferenceAlleleFromHgvs(annotation.getOriginalVariantQuery());
```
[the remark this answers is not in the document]
bprize15: After some investigation, it seems the response for VEP for dups does not return a reference allele, instead only the variant allele. For example, the variant `17:g.41242962_41242963dupTG` results in the allele_string `-/TG`. Since we cannot verify the ref allele on dup without convoluted logic, I think we can ignore for now.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VerifiedVariantAnnotationService.java
```diff
@@ -161,6 +187,10 @@ private String getReferenceAlleleFromAnnotation(VariantAnnotation annotation)
         if (slashPosition == -1 || slashPosition == 0) {
             return "";
         }
+
+        if (annotation.getOriginalVariantQuery().contains("dup")) {
+            return alleleString.substring(slashPosition + 1);
+        }
```
leexgh, line 193: Why slashPosition + 1? Do you have an example?

bprize15: For duplications, VEP only returns the ref allele. However, since it is a duplication, this is the same as the ref allele, so I just use that. For example 17:g.41242962_41242963dupTC returns allele_string -/TC so I just use the variant allele

web/src/main/java/org/cbioportal/genome_nexus/GenomeNexusAnnotation.java
```diff
@@ -67,6 +67,8 @@
 @EnableSwagger2 // enable swagger2 documentation
 public class GenomeNexusAnnotation extends SpringBootServletInitializer implements ErrorController
 {
+    // static { System.setProperty("os.arch", "i686_64"); }
+
```
leexgh, line 71:
```suggestion
```

bprize15: Ah good catch, didn't mean to commit this

PR #775, opened 2025-02-07
Allow disabling or enabling cache

leexgh:
Fix: https://github.com/oncokb/oncokb/issues/3836

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java
@@ -73,17 +73,7 @@ public abstract class BaseCachedExternalResourceFetcher<T, R extends MongoReposi
 
         // If caching is disabled, always fetch directly from the web service
          if (!cacheEnabled) {
-            try {
-                DBObject rawValue = this.fetcher.fetchRawValue(id);
-                rawValue = this.normalizeResponse(rawValue);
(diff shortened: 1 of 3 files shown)


inodb (approved):
Looks innocuous to me

service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedExternalResourceFetcher.java
```diff
@@ -67,6 +70,22 @@ protected Boolean isValidId(String id)
...
+            try {
+                DBObject rawValue = this.fetcher.fetchRawValue(id);
+                rawValue = this.normalizeResponse(rawValue);
+                List<T> list = this.transformer.transform(rawValue, this.type);
+                if (!list.isEmpty()) {
+                    return list.get(0);
+                }
+            } catch (Exception e) {
+                LOG.error("Error fetching external resource for id " + id, e);
+            }
```
onursumer, line 85: This is pretty similar to the code segment below.

https://github.com/genome-nexus/genome-nexus/pull/775/files#diff-497a870f81e19d2415bfdcbd0eb906a6894575c15bdd26b8a5fae22bab86dd9bR107-R133

Might be better to add a new method and reuse in both places.

leexgh: Moved to a new method

onursumer (approved):
LGTM!

leexgh:
Tests are failing, will try to fix in a follow up pr if it's not random

PR #777, opened 2025-02-11
Make prioritizing cancer gene transcripts configurable

leexgh:
Genome Nexus prioritizes cancer gene transcripts (oncokb transcripts) when picking canonical transcript, if people prefer to not include this prioritizer, it can be disabled in the application.properties

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/factory/IsoformAnnotationEnricherFactory.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/factory/IsoformAnnotationEnricherFactory.java
@@ -1,0 +1,34 @@
+package org.cbioportal.genome_nexus.service.factory;
+
+import org.cbioportal.genome_nexus.service.config.AppConfig;
+import org.cbioportal.genome_nexus.service.EnsemblService;
+import org.cbioportal.genome_nexus.service.OncokbService;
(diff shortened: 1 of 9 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/factory/IsoformAnnotationEnricherFactory.java
```diff
@@ -0,0 +1,34 @@
+package org.cbioportal.genome_nexus.service.factory;
+
+import org.cbioportal.genome_nexus.service.config.AppConfig;
+import org.cbioportal.genome_nexus.service.EnsemblService;
+import org.cbioportal.genome_nexus.service.OncokbService;
+import org.cbioportal.genome_nexus.service.enricher.IsoformAnnotationEnricher;
+import org.springframework.stereotype.Component;
+
+@Component
+public class IsoformAnnotationEnricherFactory {
```
leexgh, line 10: Need to add IsoformAnnotationEnricherFactory because the original IsoformAnnotationEnricher is an abstract class and not managed by Spring, so it cannot read config from application.properties, adding @Component to IsoformAnnotationEnricherFactory as this class can be managed by Spring

leexgh:
Merging now, will fix comments in follow up pr

onursumer (commented):
LGTM!

Issue #626, opened 2022-08-16
Use mskcc references as the default


inodb:
Default should be `mskcc`. We can remove `uniprot` everywhere:

- [x] genome-nexus-importer
- [x] genome-nexus-annotation-pipeline
- [x] genome-nexus
- [x] frontend can remove `uniprot`
- [x] wrapper https://github.com/cBioPortal/datahub-study-curation-tools/tree/master/GN-annotation-wrapper
- [x] wrapper https://github.com/genome-nexus/annotation-tools

leexgh:
- genome-nexus-importer: no need to change, keep uniprot files as is
- genome-nexus-annotation-pipeline: update documentation and recommend use `mskcc`
- genome-nexus: use `mskcc` as the default if no isoform override source is defined
- frontend: we don't have isoform override defined, so once genome-nexus backend uses `mskcc` as the default, the frontend should use `mskcc` automatically, no need to change anything now
- wrapper https://github.com/cBioPortal/datahub-study-curation-tools/tree/master/GN-annotation-wrapper: defaulted to using mskcc isoform across the portal already, @rmadupuri updated documentation.
- wrapper https://github.com/genome-nexus/annotation-tools: change to `mskcc` by default

PR #781, opened 2025-02-19
Use mskcc as default isoform override source

leexgh:
Partial fix: https://github.com/genome-nexus/genome-nexus/issues/626


inodb (approved):
great! Thank you!

PR #785, opened 2025-02-26
assign correct original variant query to annotation

bprize15:


service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationService.java
```diff
@@ -257,6 +257,7 @@ private List<VariantAnnotation> fetchAnnotationsExternally(List<String> variants
                         this.saveToIndexDb(normVarToOrigVarQuery.get().get(0), variantAnnotation);
                     }
                     String originalVariantQuery = normVarToOrigVarQuery.get().get(1);
+                    normVarToOrigVarQueries.remove(normVarToOrigVarQuery.get());
```
leexgh, line 260: This part is not easy to read, took me a while to understand what is get(0) get(1) and what does .remove() do. To save our future work, can you add some comments to explain it? 

For this part above, can we use a Pair or Map instead of List for `normVarToOrigVarQuery`? It's easier to read and maintain. Also good to add some comments for this part too since this is associated with the .remove() and the comments in these two parts should correspond to each other.
```
List<String> normVarToOrigVarQuery = new ArrayList<>();
            normVarToOrigVarQuery.add(normalizedVariant);
            normVarToOrigVarQuery.add(variant);
            normVarToOrigVarQueries.add(normVarToOrigVarQuery);
```

bprize15: Added a comment about the remove, also changed many of the variable names to be more explicit, so hopefully the code is more self explanatory

bprize15: Also, Pair isn't native and I don't want to add another library (or create a new data type), but I was able to make the code block you mentioned more readable by just using an array. Hoping that works.

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationService.java
```diff
@@ -224,53 +224,52 @@ private VariantAnnotation fetchAnnotationExternally(String variant, VariantType
...
+                    .filter((normToOrig) -> normToOrig[0].equals(variantAnnotation.getVariant()))
                     .findFirst();
-                if (normVarToOrigVarQuery.isPresent()) {
+                if (normalizedToOriginal.isPresent()) {
                     if (cacheEnabled) {
-                        this.saveToIndexDb(normVarToOrigVarQuery.get().get(0), variantAnnotation);
+                        this.saveToIndexDb(normalizedToOriginal.get()[0], variantAnnotation);
                     }
-                    String originalVariantQuery = normVarToOrigVarQuery.get().get(1);
+                    String originalVariantQuery = normalizedToOriginal.get()[1];
```
leexgh, line 256: Personally I still prefer map over array. I would do this with two maps instead:
1. originalVariantToNormalizedVariant map:
```
for each originalVariant in variants:
  -> get normalizedVariant = this.normalizeVariant(originalVariant, variantType);
  -> create a map {key: originalVariant, value: normalizedVariant}
```
2. normalizedVariantToVariantAnnotation map:
```
  -> get all annotations: variantAnnotations = fetchAndCache(new ArrayList<>(normalizedVariants));
  -> map each normalized variant to its annotation {kep: normalizedVariant, value: variantAnnotation}
```
3. get normalizedVariant from the first map, then get annotation from the second map, and return all annotations
```
List<VariantAnnotation> results = new ArrayList<>();
for (String originalVariant : variants) {
    String normalizedVariant = originalVariantToNormalizedVariant.get(originalVariant);
    VariantAnnotation annotation = normalizedVariantToVariantAnnotation.get(normalizedVariant);
    ...
    results.add(annotation);
}
return result;
```

Anyway, it's not a big deal, just a preference for code maintainability. I'll leave the decision to you.

leexgh (approved):
Thanks! Feel free to merge

PR #789, opened 2025-03-04
add test cases

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/671


.circleci/config.yml
```diff
@@ -13,12 +13,16 @@ jobs:
     # A list of available CircleCI Docker Convenience Images are available here: https://circleci.com/developer/images/image/cimg/openjdk
     docker:
       - image: cimg/openjdk:11.0
+      - image: genomenexus/gn-mongo:0.30
```
leexgh, line 16: Do we want to stick to this version?

leexgh (commented):
Do we need 1000+ test cases? Can we only include some general cases and corner cases? My browser is freezing on viewing this so can't load all tests lol, not sure how you picked these cases

.circleci/config.yml
bprize15: Yeah that's my plan since we need to pick a fixed version

leexgh (approved):
Thank you!

PR #791, opened 2025-03-11
fix variant and hgvsg

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/668


leexgh (approved):
LGTM. Tests seem to be failing consistently. Also, it might be good to capture this case in a test as well.

PR #846, opened 2026-04-13
Add Cancer Hotspots v3 support with opt-in API parameter

inodb:
## Summary

Adds support for Cancer Hotspots v3 data as an **opt-in** feature. Existing API consumers see no change unless they pass \`includeV3=true\`. Companion PR: genome-nexus/genome-nexus-importer#124 (adds the \`version\` column to imported hotspot data).

### What changed

- **\`model/Hotspot.java\`** — new \`@Indexed @Field(\"version\")\` string field.
- **\`web/mixin/HotspotMixin.java\`** — exposes \`version\` in JSON with Swagger docs.
- **\`service/CancerHotspotService.java\`** and **\`service/internal/CancerHotspotServiceImpl.java\`** — v3-aware overloads for all seven hotspot methods. Filtering lives in a single \`filterByVersion\` helper in the service impl. No-arg overloads are preserved and delegate to the \`includeV3=false\` path so no existing caller needs to change.
- **\`web/CancerHotspotsController.java\`** — all seven endpoints accept an optional \`?includeV3=true|false\` query parameter (default \`false\`).

### API surface

All existing endpoints are unchanged on the happy path; the new parameter is additive:

| Method | Path | New parameter |
|---|---|---|
| GET | \`/cancer_hotspots/hgvs/{variant}\` | \`?includeV3\` |
| POST | \`/cancer_hotspots/hgvs\` | \`?includeV3\` |
| GET | \`/cancer_hotspots/genomic/{genomicLocation}\` | \`?includeV3\` |
| POST | \`/cancer_hotspots/genomic\` | \`?includeV3\` |
| GET | \`/cancer_hotspots/transcript/{transcriptId}\` | \`?includeV3\` |
| POST | \`/cancer_hotspots/transcript\` | \`?includeV3\` |
| POST | \`/cancer_hotspots/proteinLocations\` | \`?includeV3\` |

### Backwards compatibility

- Default behavior (no \`includeV3\`) filters out records where \`version == \"v3\"\`. v2 records and records with a **null** \`version\` (legacy pre-v3 data) pass the filter unchanged.
- \`HotspotAnnotationEnricher\` (used in the standard variant annotation pipeline) calls \`hotspotService.getHotspots(transcript, annotation)\`, which cascades through the default path and sees v2-only data.
- \`Hotspot.equals/hashCode\` is intentionally keyed on \`hugoSymbol + residue + type + tumorCount\` and does **not** include \`version\`. If a v3 record ever collides with a v2 record on those four fields, \`Set<Hotspot>\` dedup would merge them — only matters on the \`includeV3=true\` path.

### Deploy ordering

**This PR first, importer second.** The genome-nexus-importer PR reimports the \`hotspot.mutation\` collection with 6910 v2 + 164 v3 records. If the importer runs while the old API is still live, callers would start seeing v3 records by default — breaking the contract. Once this PR is deployed, the API tolerates both the old (no-\`version\`) and new (with-\`version\`) Mongo schemas.

### Testing

Added \`testV3HotspotOptIn\` to \`CancerHotspotsIntegrationTest\`:

- Seeds two test hotspots on a dedicated transcript \`ENST99999999901\` — \`R100\` (\`version=v2\`) and \`R200\` (\`version=v3\`).
- Verifies \`GET /cancer_hotspots/transcript/{id}\` returns only R100 by default and with \`?includeV3=false\`.
- Verifies \`?includeV3=true\` returns both, with the \`version\` field populated on each.
- Verifies the same semantics on \`POST /cancer_hotspots/transcript\`.
- Backwards compatibility for null-version legacy data is covered implicitly by the existing tests (\`testTranscriptIdHotspots\` etc. use fixture records with no \`version\` field and still pass under the default endpoint).

> ⚠️ **Not locally compiled.** The development environment for this branch did not have Java/Maven available. CI compile + test is the verification gate; please do not mark ready-for-review until CI is green.

## Test plan

- [ ] CI compiles cleanly
- [ ] \`CancerHotspotsIntegrationTest\` passes, including the new \`testV3HotspotOptIn\`
- [ ] Manual smoke test against a Mongo populated by genome-nexus/genome-nexus-importer#124:
  - [ ] \`GET /cancer_hotspots/transcript/ENST00000288602\` returns the same records as today
  - [ ] \`GET /cancer_hotspots/transcript/ENST00000288602?includeV3=true\` returns the same or more records, all with \`version\` populated
  - [ ] Variant annotation (\`/annotation\` with \`fields=hotspots\`) still returns v2-only hotspots on the enricher path

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/model/src/main/java/org/cbioportal/genome_nexus/model/Hotspot.java
+++ b/model/src/main/java/org/cbioportal/genome_nexus/model/Hotspot.java
@@ -76,6 +76,10 @@ public class Hotspot
     @Field(value="splice_count")
     private Integer spliceCount;
 
+    @Indexed
+    @Field(value="version")
+    private String version;
+
     public String getHugoSymbol() {
         return hugoSymbol;
     }
@@ -162,6 +166,14 @@ public class Hotspot
 
     public void setSpliceCount(Integer spliceCount) {
         this.spliceCount = spliceCount;
+    }
+
+    public String getVersion() {
+        return version;
+    }
+
+    public void setVersion(String version) {
+        this.version = version;
     }
 
     @Override
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/CancerHotspotService.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/CancerHotspotService.java
@@ -55,4 +55,15 @@ public interface CancerHotspotService
     List<Hotspot> getHotspotAnnotationsByGenomicLocation(String genomicLocation)
         throws VariantAnnotationNotFoundException, VariantAnnotationWebServiceException, CancerHotspotsWebServiceException;
     List<AggregatedHotspots> getHotspotAnnotationsByProteinLocations(List<ProteinLocation> proteinLocations) throws CancerHotspotsWebServiceException;
+
+    // v3-aware overloads: when includeV3 is true, v3 hotspots are included in results
+    List<Hotspot> getHotspots(String transcriptId, boolean includeV3) throws CancerHotspotsWebServiceException;
+    List<AggregatedHotspots> getHotspotsByTranscriptIds(List<String> transcriptIds, boolean includeV3) throws CancerHotspotsWebServiceException;
+    List<Hotspot> getHotspotAnnotationsByVariant(String variant, boolean includeV3)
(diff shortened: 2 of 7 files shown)


leexgh:
Cancer hotspots v3 is added in another PR, close this one

PR #848, opened 2026-04-29
Fix the allele order for reversed strand

### LINKED ISSUE #847: Flip alleles order on reversed strand
ENST00000357654:c.1360_1361del returns "referenceAllele": "TC" and cannot find matching clinvar. This is because VEP returns "allele_string": "AG/-", and it's a reversed strand, based on this [pr](https://github.com/genome-nexus/genome-nexus/pull/806), AG is flipped to TC. But the allele order should also be flipped, the actual allele should be CT.

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/847
To implement a proper reverse complement: first complement each nucleotide (A <-> T, C <-> G, already fixed in https://github.com/genome-nexus/genome-nexus/pull/806), then reverse the order. For example:

 - AG - complement to TC - reverse to CT
 - ATG - complement to TAC - reverse to CAT

### DIFF
--- a/service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationService.java
+++ b/service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationService.java
@@ -439,7 +439,18 @@ public class VariantAnnotationService
 
     private void normalizeStrand(VariantAnnotation variantAnnotation) {
         if (variantAnnotation.getStrand() != null && variantAnnotation.getStrand() == -1) {
-            variantAnnotation.setAlleleString(GenomicLocationUtil.getReverseStrandAllele(variantAnnotation.getAlleleString()));
+            String alleleString = variantAnnotation.getAlleleString();
+            if (alleleString != null) {
+                String[] parts = alleleString.split("/");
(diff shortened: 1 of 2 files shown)


service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationService.java
```diff
@@ -438,8 +438,22 @@ private void addAdditionalInformation(VariantAnnotation variantAnnotation, Varia
     }
 
     private void normalizeStrand(VariantAnnotation variantAnnotation) {
+        if (variantAnnotation == null) {
```
bprize15: Do we expect variantAnnotation ever being null?

bprize15 (approved):
One small question, otherwise LGTM

service/src/main/java/org/cbioportal/genome_nexus/service/internal/VariantAnnotationService.java
leexgh: Good catch! I missed some leftover code. Deleted

PR #855, opened 2026-05-20
match variant based on id for local installation

bprize15:


leexgh (changes requested):
Current implementation will store Ensembl version of id to the Mongodb, and that becomes the key for this variant, however the key should be in hgvsg format. The next time when you call this variant again, it tries to find by hgvsg -> not found -> try to cache again

E.g. I call POST with 
```
[
  {
    "chromosome": "7",
    "start": 140453136,
    "end": 140453136,
    "referenceAllele": "A",
    "variantAllele": "T"
  }
]
```
In mongodb I got:

Image

In response I got:

Image

There are 2 issues here:
1. "_id" is the key in Mongodb, so it should be in hgvsg format, otherwise when calling the same variant again it cannot fint it.
2. Ensembl version of id should be internal and not returned in the API response, the `variant` filed in API response should be in hgvsg format


For issue 1, we can change `./service/src/main/java/org/cbioportal/genome_nexus/service/cached/BaseCachedVariantAnnotationFetcher.java` 
```
@Override
    protected String extractId(DBObject dbObject)
    {
        return (String)dbObject.get("input");
    }
```
to 
```
@Override
    protected String extractId(DBObject dbObject)
    {
        return (String)dbObject.get("id");
    }

As id has the correct hgvsg format id.

PR #867, opened 2026-07-24
Fix Entrez Gene ID resolution for HGNC-renamed genes

n1zea144:
EntrezGeneIdResolver looked up the Entrez ID using the raw VEP gene symbol, while HugoGeneSymbolResolver separately upgrades old/deprecated symbols to the current HGNC-approved one for the same transcript consequence. Since the internal symbol->Entrez map is keyed only by current official symbols, any gene affected by an HGNC rename (e.g. H3F3A -> H3-3A) resolved a correct hugoGeneSymbol but a null/missing entrezGeneId in the annotation summary, even though the Entrez ID for the current symbol was present in the data.

Reorder getTranscriptSummary() to resolve the Hugo symbol first and pass it into the Entrez lookup, which now tries the resolved symbol before falling back to the raw one.


leexgh (approved):
LGTM thank you!

inodb (approved):
LGTM - thank you!

PR #868, opened 2026-07-24
Test integration test with failed variant

leexgh:
For testing only


leexgh:
Test works as expected, close PR

PR #872, opened 2026-08-11
Fix protein position resolver on intergenic and splice region mutations

leexgh:
#### Problem:

  For splice_region_variant intronic variants (e.g. MET chr7:116411872-116411900del), Genome Nexus computes hgvspShort as p.*963* — but returns proteinPosition as null. This is
  because VEP does not provide protein_start for purely intronic positions, and the existing fallback logic in ProteinPositionResolver only handled two patterns: dup and x (for
  p.X{N}_splice). The p.*{N}* pattern was missed.

  As a result, the annotation pipeline received a null proteinPosition from the server and left the Protein_position MAF column empty.

  Note: splice_acceptor_variant / splice_donor_variant are unaffected — VEP provides protein_start for those because they carry coding_sequence_variant in their consequence terms.

  #### Root cause:

  ProteinPositionResolver.resolve() had no case for the p.*{N}* / p.*{N}fs* pattern. The extractProteinPos() helper already handles it correctly (it uses
  Numerical.extractPositiveIntegers), it just wasn't being called.

  #### Fix:

  Added a new special case in ProteinPositionResolver.resolve() — when hgvspShort matches p.*\d+.* and no position was resolved from the other cases or from VEP, extract the protein
   position from the hgvspShort string directly.

### DIFF
--- a/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/ProteinPositionResolver.java
+++ b/component/src/main/java/org/cbioportal/genome_nexus/component/annotation/ProteinPositionResolver.java
@@ -1,5 +1,7 @@
 package org.cbioportal.genome_nexus.component.annotation;
 
+import java.util.List;
+
 import org.cbioportal.genome_nexus.model.IntegerRange;
 import org.cbioportal.genome_nexus.model.TranscriptConsequence;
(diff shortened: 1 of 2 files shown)


leexgh:
I believe we should keep intergenic and splice_region mutations with null protein_position as the current c.number/3 is also not accurate, better to use null

PR #874, opened 2026-08-14
Fix hgvspShort UTR guard suppressing splice-site variants

### LINKED ISSUE #873: 3' UTR guard in ProteinChangeResolver.resolveHgvspShortFromHgvsc() also suppresses HGVSp_Short for splice-site variants
## Summary

`ProteinChangeResolver.resolveHgvspShortFromHgvsc()` returns `null` immediately whenever a
transcript's HGVSc contains `"c.*"` (3' UTR notation), before the splice-site branch that
builds the `p.X{N}_splice` synthetic label ever runs. This means `splice_acceptor_variant`/
`splice_donor_variant` variants whose relevant transcript's splice junction falls in the 3'
UTR (downstream of the stop codon) get no `hgvspShort` at all — even though the exact same
formula, in the same method, produces a correct value for other transcripts of the same
variant.

## Reproduction

```
GET https://annotation.genomenexus.org/annotation/12:g.12871890T>C?fields=annotation_summary
```

For this variant (CDKN1B, 12:12871890T>C, GRCh37):

| transcript | biotype | canonical | hgvsc | hgvspShort |
|---|---|---|---|---|
| `ENST00000228872` (RefSeq `NM_004064.3`) | protein_coding | yes | `c.*8+2T>C` | *(none)* |
| `ENST00000442489` | protein_coding | no | `c.*8+2T>C` (identical) | *(none)* |
| `ENST00000477087` | processed_transcript | no | `n.284+2T>C` | `p.X95_splice` |

Both protein-coding transcripts share the exact same HGVSc and both get no `hgvspShort`.
Only the non-coding transcript escapes — purely because its HGVSc uses `n.` notation and
never contains the literal substring `c.*`. It is not a canonical-vs-non-canonical
distinction; the canonical transcript here is one of the two blocked ones.

## Root cause

`component/src/main/java/org/cbioportal/genome_nexus/component/annotation/ProteinChangeResolver.java`,
`resolveHgvspShortFromHgvsc()`, lines ~129-133:

```java
if (m.matches())
{
    // "c.*" represents for UTR
    if (hgvsc.contains("c.*")) {
        return null;
    }
    ...
```

This guard sits above *both* downstream branches — the splice-site branch
(`p.X{pPos}_splice`, ~line 151) and the silent-mutation branch (`p.*{pPos}[fs]*`, ~line
155). Per `git log -L`, it was added in 6e68cbb ("remove hgvspShort for silent mutations",
2020-07-14). The commit message indicates the intent was to stop synonymous/silent 3' UTR
mutations from getting a bogus protein-change label — but the implementation blocks the
splice-site branch too, which is a different, unrelated case: a splice donor/acceptor site
landing just past the stop codon is a real, computable position. The formula on the very
next lines (`pPos = (cPos + 2) / 3`) produces a correct value whenever this guard doesn't
short-circuit it first — confirmed by hand for this example: `cPos=8` → `pPos=3` →
`p.X3_splice`, matching what this exact variant produced before this guard was introduced.

## Suggested fix

Scope the `c.*` guard to only the branch it was meant to protect, e.g.:

```java
if (hgvsc.contains("c.*") &&
    !SPLICE_SITE_VARIANTS.contains(transcriptConsequence.getConsequenceTerms().get(0))) {
    return null;
}
```

or move the check inside the `else if (transcriptConsequence.getAminoAcids() == null)`
branch instead of gating both.

## Impact

Any `splice_acceptor_variant`/`splice_donor_variant` whose relevant transcript's splice
junction falls in the 3' UTR gets no `hgvspShort` synthesized, even though the codebase
already knows how to compute one. Downstream consumers (e.g. cBioPortal) show these
mutations without a protein change annotation instead of the synthetic `p.X{N}_splice`
label.

n1zea144:
resolveHgvspShortFromHgvsc() bails out early for any transcript whose HGVSc contains "c.*" (3' UTR notation), before the splice-site branch that builds "p.X{N}_splice" ever runs. That guard was added to stop silent 3' UTR mutations from getting a bogus protein-change label, but it also silently blocks splice_acceptor_variant/splice_donor_variant consequences whose splice junction happens to fall past the stop codon -- a real, computable position that the very next lines already know how to derive.

Scope the guard to the branch it was meant for so splice-site variants fall through to the existing p.X{N}_splice construction instead.

Fixes #873


inodb (approved):
Thank you!

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/ProteinChangeResolver.java
leexgh, line 148: Like @inodb said `c.number / 3` would be wrong, c.*8+2 is 8 bp past the stop codon at the C-terminal end of CDKN1B but it ends up with p.X3_splice. Maybe we can do something like for 3' UTR splice site mutation, use the last protein location as the number after X. I'm not sure if it's easy to get the protein position so maybe ok to keep it as is too

leexgh (approved):

component/src/main/java/org/cbioportal/genome_nexus/component/annotation/ProteinChangeResolver.java
leexgh: Use protein length could work too, but it will need another database lookup of EnsemblService.getEnsemblTranscripts, proteinLength is stored per transcript

leexgh:
@n1zea144 @inodb I made some changes to the protein position of 3'UTR and 5'UTR splice mutations: 
- 3' UTR splice sites (c.*N±1/±2) were falling back to (cPos+2)/3, which landed the p.X{N}_splice marker in middle of the protein or N terminal even though the splice junction sits past the stop codon. I made the change to look up the transcript's protein_length via EnsemblRepository and anchor at the C-terminus.
- 5' UTR single-position splice sites (c.-N±1/±2) had the mirror-image problem: (cPos+2)/3 with the sign stripped by the regex places the marker somewhere in the middle of the protein based on how far upstream the variant is. Anchor at aa 1 (N-terminus) instead.

| Type  | HGVSc | Transcript / protein length | Before fix | After fix |
| :--- | :--- | :--- | :--- | :--- |
| 3' UTR | c.*8+2T>C | ENST00000228872 / 198 aa | p.X3_splice | p.X198_splice |
| 5' UTR | c.-100+1G>A | 100 aa | p.X34_splice | p.X1_splice |

inodb:
Thank you @leexgh !

